PR #3, opened 2015-01-29
Change color in editor: Grey -> Gray

gus-massa:
I'm not sure that this is an error, because I couldn't trigger it. 
 
I suspect that this "Grey" color is ignored or silently transformed to RGB=#000. The other colors in the list are in the default color database, but this isn't.


mflatt:
merged - thanks!

PR #4, opened 2015-03-04
Use object=? instead of eq? to handle equality for contracted objects

dfeltey:
This pull requests changes several instances of `eq?` in the pasteboard implementation to instead use `object=?` allowing for contracted objects to be compared, and improving interop with typed racket programs using pasteboards.

### DIFF
--- a/gui-lib/mred/private/wxme/pasteboard.rkt
+++ b/gui-lib/mred/private/wxme/pasteboard.rkt
@@ -237,7 +237,10 @@
                              ;; find snip:
                              (let ([snip (find-snip x y)])
                                (and snip
-                                    (eq? snip s-caret-snip)
+                                    (or (and (object? snip)
+                                             (object? s-caret-snip)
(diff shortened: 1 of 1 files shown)


samth:
It looks like this might be simplified w/ a function that did the `object?` tests, rather than duplicating that a bunch of places.

dfeltey:
I agree, but should that function go here or in the class implementation instead, since there is some of the same sort of duplication in my snip pull request as well.

mflatt:
I think I'd rather have more specific tests. Often (probably always, but I'll have to look more closely), the alternatives are an object or #f, and using a combination of `object=?` and checking for #f seems locally clearer.

I'll take a closer look and merge some variant of the patches.

mflatt:
inline checks for `#f` do not read well; adding `object-or-false=?` seems best to me, so far

mflatt:
merged

PR #5, opened 2015-03-09
Delete Trailing whitespace on return

maxsnew:
This deletes trailing whitespace from the current line when you press return. Should eliminate a lot of accidental whitespace.

I also added tests for `insert-return` with the ability to specify `tabify-on-return?`, but I didn't add any tests where `tabify-on-result?` was `#f` since I couldn't figure out test cases that actually depended on that value.


rfindler:
I am not seeing any code that changes the value of `tabify-on-return?` so this is code that isn't tested and probably isn't used.

But, that said, it is a relatively easy piece of functionality to keep, so I'm in favor of keeping it.

maxsnew:
Ok I've fixed that, added tests for when `tabify-on-return?` is `#f` and squashed everything to a single commit.

rfindler:
Okay, I've pushed the commit. 

Thanks!

PR #8, opened 2015-04-21
gtk support for frame% fullscreen and is-fullscreened?

jkominek:
Not much to it. All the pieces were already in place.


mflatt:
Merged - thanks!

PR #10, opened 2015-04-27
win32 support for frame% fullscreen and is-fullscreened?

jkominek:
This has a bit more to it than the gtk fullscreen commit (#8), which is why I left it separate.

I referred to, and basically translated, the way that GTK performs fullscreen/unfullscreening on win32:
https://github.com/GNOME/gtk/blob/bbac0eb3b938769d450310022ab7d5e8109525e4/gdk/win32/gdkwindow-win32.c#L2813

tested on a three monitor win7 64bit machine, as that's all i can test this on.

happy to do any followup work that it needs.


mflatt:
Works for me, too, so I've merged the change - thanks!

PR #12, opened 2015-05-20
Change some `eq?`s in snip handling code.

samth:


rfindler:
I'm not sure why we should change an `object=?` into an `eq?`. Maybe lets do the ones that go the other way?

samth:
Changed to just the ones that go the appropriate way.

rfindler:
Let's merge this after branch day.

PR #13, opened 2015-06-15
Fix typo, clarify get-file filter

lehitoskin:
The function get-file-list had a typo in its description.

Also, the paragraph about the filter in get-file was a little vague about how "simple" the pattern blob could be, so I attempted to expound and clarify just how simple it is and what a, example filter string could look like.


stamourv:
Merged, thanks!

PR #14, opened 2015-08-03
Use a regular expression to indent for loops in framework

lexi-lambda:
The list of for loops is growing quite long, and as I add more, I'm beginning to grow a little bit frustrated that I need to add each new loop form to the list. Furthermore, there is no way to distribute these indentation settings with my package, so every user would need to set them for themselves.

This simply adds the regex `^for\*?(/|$)` to include all for loops, both built-in and user-defined. With this change, manually including the zoo of for loops in the indentation list is unnecessary, cleaning it up considerably.


rfindler:
Thanks!

PR #17, opened 2015-10-04
Improve scrolling with certain touchpads under Win32

bmastenbrook:
See http://msdn.microsoft.com/en-us/library/windows/desktop/dd562171(v=vs.85).aspx
- the touchpad on the Surface Pro 3 (and probably other "precision touchpad" devices) generates SB_THUMBPOSITION events. This change forces a display refresh when receiving those events. However, touch scrolling events still cause corruption due to missed redraws, and the over-scroll indication does not seem to be working properly.


bmastenbrook:
@mflatt Should I close this now that it's been merged?

PR #18, opened 2015-10-22
Add length parameter to call of `string-snip% insert`

thilliker:


stamourv:
Thanks for the fix!
According to the `string-snip%` docs, this length argument is mandatory, which would mean that this code currently cannot possibly work. Is that correct?
If so, do you have a program that triggers that faulty behavior which we could add as a test case?

rfindler:
```
#lang racket/gui
(send the-clipboard set-clipboard-string "x" 0)
(send (new pasteboard%) paste)
```

(but it screws with your clipboard -- there is probably a better test case)

thilliker:
Yes, the current code cannot possibly work. 
I found this error using ctrl+v in a pasteboard% window. The program which alerted me to this simply uses a pasteboard% & doesn't do any programmatic copy/paste.

stamourv:
Merged, thanks!

PR #19, opened 2015-10-31
Swap get-end-position and get-start-position 

LeifAndersen:
(They were backwards in the docs. I'm not sure if that was intentional or not.)


rfindler:
Thanks!

PR #25, opened 2016-02-03
Parenthesis color scheme default to "Spring"

gfbee:
Change default to one that highlights nested structure, i.e. anything but the current "Basic grey".
As the default for a new user, "Spring" has these advantages:
   - lightest (versus "Shades of grey" or "Fall")
   - emphasizes outer structure more than inner (e.g. versus "Winter")
   - doesn't suggest selection (e.g. versus "Shades of blue"), instead it suggests yellow highlighting

### DIFF
--- a/gui-lib/framework/private/main.rkt
+++ b/gui-lib/framework/private/main.rkt
@@ -85,7 +85,7 @@
                                     l))
                              values)
 
-(preferences:set-default 'framework:paren-color-scheme 'basic-grey symbol?)
+(preferences:set-default 'framework:paren-color-scheme 'spring symbol?)
   
 (define cond/offset-defaults
   '(("case-lambda" 0)


rfindler:
This one has performance implications and changing this default will affect
all kinds of things, e.g., TA instructions and instructor handouts. So
changing it seems unwise to me.

If you wanted to make it easier to find the preference setting somehow via
some other change that seems possible tho. Any ideas?

gfbee:
Nothing thought out, but either a splash screen when DrRacket first runs, the way many apps now mention common out-of-the-box information. Or something where we already do it: when telling the user that a language must first be chosen.

It seems a shame that a few defaults that are so commonly set and make things so much nicer are missed by many people. I inform my students in various ways, but the beginners don't do it for various reasons, and the experienced don't do it for other reasons (although with some overlap of reasons).

Eventually when racket becomes mainstream these tips will appear in many places and become part of the culture, the way the folklore for other languages does, but for now that doesn't appear to be the case.

rfindler:
Do you have a way to measure how commonly various defaults are changed?

gfbee:
Good point, I should be more careful and precised in a discipline that includes the word "Science".
What I've observed over the years is whenever we have a student change the highlighting, I don't recall a single instance where they didn't say the liked it. This has happened outside of the context of a class as well. I only have a moment now, and I forget the exact wordings and combination of features for insertion of parentheses, but people I've known tend to prefer the automatic balancing and context-dependent automatic conversion of brackets to parentheses.

rfindler:
How is this for a deal: if you run a poll on users for these three
preferences and get a good number of people responding, then I'll change
the way they say they use it? Maybe other instructors would also circulate
the poll with their classes too?

Or is that too much work?

gfbee:
That's exactly what I was thinking of doing, just didn't want to mention it until I thought through the details and could commit to it.

rfindler:
Cool!

PR #27, opened 2016-02-06
Add with-* to the set of lambda like expressions.

LeifAndersen:


LeifAndersen:
Correct, no need to make a merge commit. I'll merge this now.

LeifAndersen:
merged.

PR #31, opened 2016-05-12
Porting parts of gui-lib to Typed Racket

samth:
This is an initial pull request -- not ready to merge yet.

Work done by @earldean.

### DIFF
--- a/0001-gui-utils.rkt-ported-to-typed-racket.patch
+++ b/0001-gui-utils.rkt-ported-to-typed-racket.patch
@@ -1,0 +1,405 @@
+From 30df05652395718c82c59df6a9d92702c70dcb7b Mon Sep 17 00:00:00 2001
+From: Earl Dean <earldean@indiana.edu>
+Date: Wed, 21 Oct 2015 18:22:21 -0400
+Subject: [PATCH] gui-utils.rkt ported to typed racket
+
+---
+ gui-lib/framework/gui-utils.rkt | 159 ++++++++++++++++++++++++++++++----------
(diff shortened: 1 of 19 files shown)


gui-lib/framework/private/Untitled.rkt
```diff
@@ -0,0 +1,7 @@
+#lang racket
+
+(define specs null)
+
+(set! specs (cons (list "hello" "world") specs))
+
+(cons (list "hello" "world") specs)
```
stamourv: I think this file was included by mistake.

0001-gui-utils.rkt-ported-to-typed-racket.patch
```diff
@@ -0,0 +1,405 @@
+From 30df05652395718c82c59df6a9d92702c70dcb7b Mon Sep 17 00:00:00 2001
+From: Earl Dean <earldean@indiana.edu>
+Date: Wed, 21 Oct 2015 18:22:21 -0400
+Subject: [PATCH] gui-utils.rkt ported to typed racket
```
stamourv: I think those got added by mistake.

gui-lib/framework/private/Untitled.rkt
samth: Yeah, this is not worth reviewing yet.

stamourv:
The main concern I have is: will this affect the usability of the GUI system from untyped Racket? I'm worried about a repeat of the math library situation.

samth:
I don't think we plan to merge this or keep working on it.

PR #48, opened 2016-09-22
Adds a preferences list for identifiers which behave like ellipses

jsmaniac:
The indentation behaviour of the `...` ellipses was hardcoded, so `...+` would not indent as expected, nor would the unicode ellipses "…" and "…+"

```
#'(hello ...
   world)
#'(hello ...+
         world)
```

With this patch, a new list is added to the preferences dialog, along with the lists of lambda-like keywords, define-like keywords, etc.

The list can then be customized, and by default it includes the following symbols:

```
#'(hello ...
   world)
#'(hello ...+
   world)
#'(hello …
   world)
#'(hello …+
   world)
#'(hello ::...
   world)
```


rfindler:
This seems like a good idea. Three concerns:

1) what is the new minimum width of the preferences dialog? (It has to be kept narrow)

2) user's old preferences will be clobbered by this change. The simplest fix is probably to not change the tabify preference but instead add a new one. If you were to change the tabify pref, then you'd have to compensate by adding an unmarshaller that used the old prefs and updated to the new one (by adding in the defaults).

3) where did you get that initial list from? I've seen `...+` but not the others.

jsmaniac:
@rfindler Good questions.

1) The new window is 1133×485 The old window would have been roughly 222 px less, so roughly 911×485 .

2) What exactly do you mean by "clobbered"? The new code checks whether the tabify pref has four or five lists, and uses a default value for the fifth in the former case. After DrRacket is restarted (or after the user goes at least once in the preferences dialog?) the new five-list version is saved. The other lists kept their contents (I didn't have to re-add identifiers to the lists), and the new one got initialized with the defaults as you would expect. I see one potential problem here: maybe the temporary default is just an empty list, I didn't check how it behaves when using the new version for the first time.

3) `...` is from racket/base, `...+` from syntax/parse, `::...` from math/array, and the unicode variants `…` and `…+` are from my personal toolkit phc-toolkit, but I think Jay mentionned durin RacketCon that unicode ellipses are part of `#lang remix`, so I thought it would be good to include them by default and not just add them on my own system. I can remove the last two and/or the `::...` if you think it's better to not put them in by default.

rfindler:
1) that seems too wide to me. How about making two panes, with 3 in one pane and 2 in the other?

2) Oh, I missed that. But that means that all code that uses that preference (i.e. passes the symbol to `preferences:get`) has to do this check. Another way to handle this is to use `preferences:set-un/marshall` (or, as I mentioned, use two preferences).

3) thanks, I agree that you've got the right default (with all of them).

jsmaniac:
@rfindler Thanks for the feedback!

You're right, it's too wide for e.g. 1024x768 and there are quite a few embedded-ish systems with this kind of lower resolution (e.g. my OpenPandora pocket computer has 800x400, and racket otherwise runs fine on it, albeit slowly).

Right now I don't have the time to fix the width and preferences (I have never used racket/gui nor the framework, so I have to look things up a bit), so I'll leave this stalled and come back to it in a couple of months.

rfindler:
Thanks for your effort and I'll keep it on my list to look into too.

Issue #23, opened 2015-12-05
Scrolls too fast with 2-finger scrolling on laptop


samth:
With the new `racket/gui` scrolling, things scroll very fast on my laptop when using 2-finger scrolling on the touchpad (much faster than in Chrome or Firefox on the same laptop, or with the mouse wheel in DrRacket on a different machine).

mflatt:
@samth Has this problem been fixed for you by Will's patch?

samth:
Yes, this is now fixed.

PR #55, opened 2016-11-04
GTK+: accumulate small scroll events into appropriate wheel events

willghatch:
This fixes the crazy fast scrolling that has been present on GTK since commit 2c5b1480f4f7b39e84548d2a618e30525178a364.


willghatch:
Oh, and now that I look through the issue list as well, this is a fix for issue #23.

PR #61, opened 2016-12-13
add replace-all check to see whether found txt is deletable

### LINKED ISSUE #60: infinite loop on `replace-all` in drracket
Found another issue while testing #59 (but unrelated to #59).

DrRacket goes into inf loop:
1) start with empty buffer except for "#lang racket"
2) press run
3) with focus on interactions window, open search box, enter "a" as search term
4) choose Edit -> Replace All
5) inf loop

Investigating now.

stchang:
- fixes #60

### DIFF
--- a/gui-lib/framework/private/frame.rkt
+++ b/gui-lib/framework/private/frame.rkt
@@ -2538,13 +2538,15 @@
                                                                         case-sensitive-search?
                                                                         #t)])
                 (when found-pos
-                  (unless (hash-ref ht found-txt #f)
-                    (hash-set! ht found-txt #t)
-                    (send found-txt begin-edit-sequence))
                   (let ([start (- found-pos (send find-edit last-position))])
-                    (send found-txt delete start found-pos)
-                    (copy-over replace-edit 0 (send replace-edit last-position) found-txt start)
-                    (loop found-txt (+ start (send replace-edit last-position)))))))
+                    (cond [(send found-txt can-delete? start (- found-pos start))
+                           (unless (hash-ref ht found-txt #f)
+                             (hash-set! ht found-txt #t)
+                             (send found-txt begin-edit-sequence))
+                           (send found-txt delete start found-pos)
+                           (copy-over replace-edit 0 (send replace-edit last-position) found-txt start)
+                           (loop found-txt (+ start (send replace-edit last-position)))]
+                          [else (loop found-txt found-pos)])))))
             (hash-for-each ht (λ (txt _) (send txt end-edit-sequence)))))))
                              
     (define/private (pop-all-the-way-out txt)


rfindler:
I don't see how this is related to #60 and, indeed, I also can't seem to get the instructions in #60 to cause a loop.

Specifically, when I make this change to frame.rkt (the version without your changes):

```
--- a/gui-lib/framework/private/frame.rkt
+++ b/gui-lib/framework/private/frame.rkt
@@ -2529,6 +2529,7 @@
             (hash-set! ht txt #t)
             (let loop ([txt (pop-all-the-way-out txt)]
                        [pos 0])
+              (printf "loop: ~s ~s\n" (eq-hash-code txt) pos)
               (let-values ([(found-txt found-pos) (find-string-embedded txt
                                                                         search-str
                                                                         'forward
@@ -2545,7 +2546,8 @@
                     (send found-txt delete start found-pos)
                     (copy-over replace-edit 0 (send replace-edit last-position) found-txt start)
                     (loop found-txt (+ start (send replace-edit last-position)))))))
-            (hash-for-each ht (λ (txt _) (send txt end-edit-sequence)))))))
+            (hash-for-each ht (λ (txt _) (send txt end-edit-sequence))))))
+      (printf "done\n"))
                              
     (define/private (pop-all-the-way-out txt)
       (let ([admin (send txt get-admin)])
```

and then I put "a" in the search field and "b" in the replace field and then do "replace all" with the focus in the interactions window, I see this output:

```
loop: 1477422 0
loop: 1477422 15
loop: 1477422 69
loop: 1477422 73
loop: 1477422 79
done
```

I also tried with "et" in search and "b" in replace (for something longer in search that has more that one hit) and "a" in search and something long in replace. No infinite loops in those cases, either.

Can you say more clearly why you think this commit is changing something from being an infinite loop? And perhaps say more precisely what state you had DrRacket in, now that you've identified some code to change?

stchang:
That's weird. I added the same printf and mine printed 15 repeatedly.

I had an empty replace box. Maybe that is the difference?

The problem was that it would not advance the `pos` properly if the text to delete was undeletable (like the "Welcome to DrRacket ..." in the interactions window).

rfindler:
Ah yes, I see what you mean. (Replace-all shoudn't be enabled when there is an empty replace string tho, right?)

Calling `can-delete?` isn't right; that's a method that supposed to be called only by `delete`.

I think a better approach is to check the revision number before and after the delete. Or maybe before and after the delete+copy. If the number changed, then we can assume the replace worked. If the number didn't change, then it didn't. 

I'll make that change.

rfindler:
Thanks. I did this a slightly different way than in this pull request (and also added some tests)

rfindler:
Here's the commit: https://github.com/racket/gui/commit/5fed92d72a6bb31b52ffa8ae177a0e74e9c8e38c

stchang:
> Replace-all shoudn't be enabled when there is an empty replace string tho, right?

No, it should be enabled. I disabled it when the *search* box is empty, but you should be able to replace a non-empty search string with nothing, right?

mfelleisen:
> On Dec 14, 2016, at 2:28 PM, Robby Findler <notifications@github.com> wrote:
> 
> Ah yes, I see what you mean. (Replace-all shoudn't be enabled when there is an empty replace string tho, right?)


Of course it is useful! I often use this to delete text from a buffer, especially when I don’t want to leave behind some whitespace. 

— Matthias

stchang:
Commit 5fed92d looks good, except I wasn't sure if `ht` should be updated if no replacement is made?

Thanks for fixing my fix.

rfindler:
The `ht` tracks which editors are in edit sequences. So if we didn't add it, then we wouldn't close the edit sequence. We have to put the editor into an edit sequence before we know if the deletion succeeds (because we don't want to see the delete until the insertion happens etc)

And right: good point about replacing things with the empty string!

stchang:
Ok, makes sense. Thanks for the explanation.

rfindler:
Thanks very much for your work on this! I wouldn't have been able to do
what I did it without it.

PR #62, opened 2016-12-15
document non-empty str requirement for find-string methods

stchang:


rfindler:
How about using `non-empty-string?`?

stchang:
Yes, that's better.

stchang:
Ok I updated and squashed

rfindler:
Thanks!

PR #69, opened 2017-03-02
Fix a subtle bug in the move/copy-to-edit method when the source and destination objects are the same

dfeltey:


dfeltey:
I'm closing this PR because I ended up adopting a different strategy to fix move/copy-to-edit.

PR #75, opened 2017-04-29
Pin the menu object while it is displayed, so that it is not garbage-collected

### LINKED ISSUE #74: Glitch: right-click menu disappears immediately
_From @jsmaniac on September 7, 2016 17:41_

The right-click menu in DrRacket sometimes disappears instantly, or sometimes appears as a small rectangle a few pixels high (ususally, it is briefly displayed correctly then turns into the tiny rectangle).

It happens on my computer (NixOs Linux 64-bits, Enlightenment 16 desktop environment), but I also saw it happen on Windows.

Here's a video showing the bug: https://vimeo.com/181820735 (zipped video here for achival purposes: [bug.mkv.zip](https://github.com/racket/drracket/files/459902/bug.mkv.zip))

With small files the bug does not occur very often (about 10% of the time), but on larger files (or maybe with more tabs open?) it occurs more frequently, and sometimes I have to do a dozen right-clicks before I can access the menu. This bug is not something new, it's been there for several months at least.

_Copied from original issue: racket/drracket#72_

jsmaniac:
Fixes #74.

It's not the cleanest fix, I guess :-( but it's short enough, and solves a very irksome bug.

Hopefully this does not create any memory leaks — is it possible for the menu to disappear without calling neither `do-selected menu-item` nor `do-no-selected` ? Also, should multi-threading issues be taken into account? Racket does not normally allow sharing mutable data between threads in an unsafe way, so I suppose this implementation should be okay without the need for further caution.

### DIFF
--- a/gui-lib/mred/private/wx/gtk/menu.rkt
+++ b/gui-lib/mred/private/wx/gtk/menu.rkt
@@ -76,6 +76,10 @@
     (define/public (removing-item) (void))
     (super-new)))
 
+;; Globally reference any opened menu, so that the menu object stays alive
+;; while it is displayed, even if no other code refers to it.
+(define global-prevent-gc (make-hasheq))
+
 (defclass menu% widget%
   (init label
         callback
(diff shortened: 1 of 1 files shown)


mflatt:
This seems like a good repair, but I don't see the next for the extra key. Will mapping `this` to `#t` in the hash table work just as well?

jsmaniac:
@mflatt Good thinking, the code is much simpler without the key. I didn't think it through, but of course the identity of `this` won't change across time :) . I pushed an update.

PR #93, opened 2018-01-09
Fix bug where DrRacket could loose data when saving.

LeifAndersen:
At the moment if a user refuses to save a file with non-textual widgets
using wxme, DrRacket will just save the file as flattened text. This leads
to lost data, and makes otherwise valid racket files invalid.

This patch prevents the file from loosing that data on saving. Namely,
if a user:
1. Opened a file as a text file.
2. Added non-textual snips to the file.
3. Has in their preferences to warn before changing the file type.
4. Saves the file but refuses to change the type.

then DrRacket will not save the file because it cannot
do so.

### DIFF
--- a/gui-lib/framework/private/text.rkt
+++ b/gui-lib/framework/private/text.rkt
@@ -2295,31 +2295,40 @@
           [(is-a? s string-snip%)
            (loop (send s next))]
           [else #f])))
+
+    ;; Saving in text when wxme is needed looses data,
+    ;; therefore if the user refuses to change file formats
+    ;; abort save.
+    (define/augment (can-save-file? name format)
+      (define needs-wxme?
+        (and (not (all-string-snips))
+             (eq? format 'same)
+             (eq? 'text (get-file-format))))
+      (define proper-format
+        (cond [(not needs-wxme?) #t]
+              [(and needs-wxme?
+                    (or (not (preferences:get 'framework:verify-change-format))
+                        (gui-utils:get-choice
+                         (string-constant save-in-drs-format)
+                         (string-constant yes)
+                         (string-constant no)
+                         #:dialog-mixin frame:focus-table-mixin)))
+               (set-file-format 'standard)
+               #t]
+              [else #f]))
+      (and proper-format (inner #t can-save-file? name format)))
     
     (define/augment (on-save-file name format)
-      (let ([all-strings? (all-string-snips)])
-        (cond
-          [(and all-strings?
-                (eq? format 'same)
-                (eq? 'standard (get-file-format))
-                (or (not (preferences:get 'framework:verify-change-format))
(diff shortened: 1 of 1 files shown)


rfindler:
I am not sure this is a better state that before. In particular, while I do agree that guis where the user can make a bad decision and then lose data need fixing (like this situation you've found), this particular fix still has the problem that the user can make a bad decision and then lose data (because they may modify the file and then lose their edits because they weren't saved, which might represent lots of work).

Perhaps a better way would be to change the wording in the dialog so that it offers the choice between "Do not save the file" and "lose information in the saved file", with a button in the dialog that sends them to the documentation in a place that explains the issue in more detail?

LeifAndersen:
Changing the dialog would be okay. I just didn't because I can't really do the translations for other languages.

For what its worth, it doesn't just quietly not save your file. If you, for example, try to quit, you'll still get a message that your file isn't saved. But I do agree that even this is not ideal.

LeifAndersen:
Shall I try the dialog change approach you suggested?

rfindler:
That would be great if you'd give that a try! Thanks.

Don't worry about translating, there are volunteers that notice new strings appearing (in the string-constants pkg) and do the translations at their own speeds. Just be sure there are new string constants (as long as there are new strings, of course, reuse existing ones as appropriate).

rfindler:
(and yes, I understand what the can-*? methods do; I meant "at that moment ...".)

LeifAndersen:
I should note that this change coincides with https://github.com/racket/string-constants/pull/21

rfindler:
PS: when you do change the string-constant files, the versions need updating.

rfindler:
I think the buttons should be more meaningful, not just "yes", "no" and "cancel". Those three used to be all the rage, but they can be super confusing in situations like this. Things have, in recent years (decades?) moved to where buttons have their dialog-specific words on them and that's been good, IMO.

LeifAndersen:
That works for me. Perhaps it could be something like: "Keep plaintext", "Convert", and "Don't Save"?

rfindler:
I like "Convert" and "don't save".  I was trying to think of a variant of the first button's string that signals "you will lose information", but am not coming up with anything.

LeifAndersen:
I agree. The best I've got is `Convert (Recommended)`, or `Keep (may loose data)`.

rfindler:
I like it!

rfindler:
(Except lose one of the "o"s to make "lose". )

LeifAndersen:
Okay, will do.

LeifAndersen:
Okay, pushed. I've also tested it on my machine and it seems to work and I'll go squash+merge these changes once the tests pass if it looks good to you.

rfindler:
Sounds good. Thanks!

Issue #94, opened 2018-02-01
calling `register-collecting-blit` leaves a blit around on macOS


florence:
It looks like each call to "register-collecting-blit" causes one permanent blit to appear. This blit acts a separate tiny window in expose mode, even after the blit is unregistered.

Example program:
```
#lang racket/gui
(define f (new frame% [label ""]))
(define hp (new horizontal-panel% [parent f]))
(define c (new canvas% [parent hp]))
(define on (make-bitmap 20 20))
(define off (make-bitmap 20 20))
(send c min-width 20)
(send c min-height 20)
(define bdc (make-object bitmap-dc% on))
(send bdc set-smoothing 'smoothed)
(send bdc clear)
(send bdc set-bitmap off)
(send bdc set-smoothing 'smoothed)
(send bdc set-brush "black" 'solid)
(send bdc set-pen "black" 0 'transparent)
(send bdc draw-ellipse 0 0 20 20)
(send bdc set-bitmap #f)

(register-collecting-blit c 0 0 20 20 on off)
(send f show #t)

(for ([i (in-range 4)])
  (unregister-collecting-blit c)
  (register-collecting-blit c 0 0 20 20 on off))
```

This causes 5 new blit "windows" to appear, sometimes with the on bitmap, sometimes with the off bitmap.

This has been causing a bit of a problem w/ DrRacket because each new window creates a new blit. After a while all the extra blits cause a noticeable slow down in expose mode.

I'm on macOS 10.13.2, and Racket v6.12.0.3.

jbclements:
Yes! I believe this is the same issue that I reported in my Dec. 20 message to racket-users, when I wrote:

> Looks to me like DrR is leaking… windows.

> 
> Specifically:
> 
> On up-to-date DrR (just rebuilt from repo, version 6.11.0.5--2017-12-20), pretty much up-to-date OS X (10.13.1 (17B1003)), I notice something very strange. I’ve seen it on two computers, so I think it should be reproducible on any computer:
> 
> 1) Open a DrR Window, nice and big.
> 2) Grab the right edge of the window, and slowly drag it to the left, shrinking the window horizontally. Watch the area of the GC box.
> 3) You should see a little trail of white left behind the lower right corner of the window.
> 4) After about 1.5 seconds of slowly dragging, let go.
> 5) The trail is still there. What is it?
> 6) to find out, use a three-finger upward wipe; this separates all of the windows on the screen, so you can see them separately. Here’s what it looks like, for me:
> 
[...]

PR #99, opened 2018-03-22
Fix blit windows hanging around when suspended

Bubba:
`removeChildWindow` doesn't actually hide the blit window so call `orderOut:` on it as well.
Fixes issue #94


mflatt:
Thank you! For whatever reason, I still haven't been able to replicate the problem. I hope your repair works for others, too.

PR #100, opened 2018-03-22
make scribble double click select words, not lines

florence:
Currently, in at-exp languages, a double click on text selects the line not the word. This changes that.


rfindler:
Thanks!

PR #102, opened 2018-03-26
factor expandable-snip etc out of syntax-browser snip

rmculpepper:
Adds two new snip classes:
- `arrow-toggle-snip%` is a toggle control. Uses include syntax snips and hierlist lists. This PR changes the snip to scale with the style's font size.
- `expandable-snip%` extracts the toggle-for-details part of syntax snips.


rfindler:
Looks good to me.

PR #106, opened 2018-04-17
graphicsContextWithGraphicsPort -> graphicsContextWithCGContext

LeifAndersen:
The former has been deprecated and replaced with the latter.


mflatt:
It looks like `graphicsContextWithCGContext` exists only in 10.10 and up, so this change would need to be conditional on the OS version (since we aim for 10.5 and up). There are functions like `version-10.10-or-later?` in "utils.rkt", but probably it's ok to just leave this alone.

LeifAndersen:
Ah, fair point. I agree its not worth it.

PR #109, opened 2018-05-21
improve KMP in wxme\text

gus-massa:
1) Check that `m` is small enough instead of waiting until the last character to exit. (This is useful for long `word`s.)

2) When a `#f` is found in the kmp-table, advance the position of `m` to the current position `m+i` instead of just advancing it once to `m+1`.

3) I'm worried about the 0 length string. I'm not sure if the result when `word=""` is `#t` or `#f` or what. I guess that case is filtered somewhere, but I broke that corner case in my first version of racket/string.rkt.

4) I think that the `(loop (+ m 1) 0))` in the lines 3790 and 3797 can be changed also to something like the code in like 3800 `[else (define t-i ...) ...)`  but I'm not sure if the kmp table needs an special entry for that case or the last entry can be used. I still have to think more the details of this.

I'm still worried that these changes break something of the multiple overlapping results.

Also, I didn't run this, so try it a few times before merging.

### DIFF
--- a/gui-lib/mred/private/wxme/text.rkt
+++ b/gui-lib/mred/private/wxme/text.rkt
@@ -3808,22 +3808,30 @@
          (if just-one? #f '())])))
   
   (define/private (build-table word)
-    (define t (make-vector (string-length word) #f))
-    (when ((string-length word) . > . 1)
-      (vector-set! t 1 0)
-      (let loop ([pos 2]
+    (define l (string-length word))
+    (define t (make-vector (+ l 1) #f))
+    (when (l . > . 0)
+      (let loop ([pos 1]
                  [cnd 0])
-        (when (< pos (string-length word))
-          (cond
-            [(char=? (string-ref word (- pos 1))
-                     (string-ref word cnd))
-             (vector-set! t pos (+ cnd 1))
-             (loop (+ pos 1) (+ cnd 1))]
-            [(> cnd 0)
-             (loop pos (vector-ref t cnd))]
-            [else
-             (vector-set! t pos 0)
-             (loop (+ pos 1) cnd)]))))
+        (cond
+          [(pos . = . l)
+           (vector-set! t pos cnd)]
+          [(char=? (string-ref word pos)
+                   (string-ref word cnd))
+           (vector-set! t pos (vector-ref t cnd))
+           (loop (+ pos 1) (+ cnd 1))]
+          [else
+           (vector-set! t pos cnd)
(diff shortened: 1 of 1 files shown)


rfindler:
I think there is something going on with embedded editors. I didn't get far enough to figure out what is actually going wrong, just far enough to find a counterexample:

```
#lang racket/gui
(define t1 (new text%))
(define t2 (new text%))
(send t2 insert "abaabccc")
(send t1 insert (new editor-snip% [editor t2]))
(send t1 find-string-embedded "abcc" 'forward 0)
```

After the commit, this returns `#f`, but before it returns the embedded editor and `3`.

gus-massa:
The size of the subeditors is not counted in the length of the container, so it's wrong to cut the search when the remainder part looks too short. I.E. the change in line [L3783](https://github.com/racket/gui/pull/109/commits/a1630c65862563b1c43d7a88905cf16dfb6c7792#diff-62180fd67008808e32b6dca8ff11815aL3783) is wrong. I'll try to fix the other PR first, and then fix this.

gus-massa:
This is **untested**, but it looks good. (I get an error because I'm trying to run it in the installed version of Racket in my machine, so I get "_cannot instantiate `racket/gui/base' a second time in the same process_".)

I think that now each position that is an embedded editor is checked only once. (A character may be checked twice when it's not the expected character.) In case this works, I'd remove the lines commented with `#;` in the last commit and squash all of them.

rfindler:
This doesn't quite seem to be right yet. There are a bunch of test failures I see; this one is the simplest, I think:

```
#lang racket/gui
(define t1 (new text%))
(send t1 insert "yyx")
(send t1 find-string "yx" 'forward 0)
```

It produces `#f` instead of `1`.

gus-massa:
I think I fixed it. It was an error in the generation of the kmp table.

gus-massa:
I squashed everything in two commits, but I transferred the additional tests to https://github.com/racket/gui/pull/112. 

A big part of the apparent changes is due to an indentation change because I removed a `case`, because it's not necessary to use `embedded-editors-considered`. Something like 

      (case 
        [(x embedded-editors-considered) y]
        [else (long-z)])
    ==>
       (long-z)

I think that this PR is correct, but it's tricky, so I prefer to wait a few days until the 7.1 branch is created, and left this PR to be merged in 7.2 with more time to test.

rfindler:
I gave it a try and didn't spot any problems. Thanks for working on this!

PR #111, opened 2018-06-08
macOS: hide the fake root frame when it becomes main

jamesbornholt:
We already hide the fake root when it's first designated
(in designate-root-frame), but macOS can make it visible
again when no other windows exist and the application is
reactivated. So hide it in response to windowDidBecomeMain.

Fixes racket/drracket#202.


mflatt:
Thanks!

Issue #82, opened 2017-10-27
Tests expect to be run in a particular directory


samth:
Running `racket -l tests/gracket/test` appears to only work in the `tests/gracket` directory. Otherwise you get an error about finding `editor.rktl` in the current directory.

Issue #39, opened 2016-08-12
reindent-paragraph incorrect behaviour with long lines


jsmaniac:
In the following code, the line `bbbb…` is two caracters longer than the maximum line length for `reindent-paragraph`. It is preceded by a shorter line `aa`.

```
#lang scribble/base

aa
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
Hello world lorem ipsum hello world lorem ipsum hi world
lorem ipsum hello world lorem ipsum hi world lorem ipsum
hello world lorem ipsum hello world lorem ipsum hello…
```

When pressing <kbd>Alt</kbd>+<kbd>Q</kbd>, the third line is appended to the `bbbb…` one, which results in a line way longer than the maximum:

```
#lang scribble/base

aa
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb Hello world lorem ipsum hello world lorem ipsum hi world
lorem ipsum hello world lorem ipsum hi world lorem ipsum
hello world lorem ipsum hello world lorem ipsum hello…
```

 If <kbd>Alt</kbd>+<kbd>Q</kbd> is pressed again, the following line is appended, and so on. The `reindent-paragraph` should instead leave `bbbb…` alone on its line (in my case it was a long `@hyperlink` causing this behaviour).

Tested with Racket 6.6.0.3.

rfindler:
This was fixed by this commit: https://github.com/racket/gui/commit/0ae0a5b8044faecb49cc10b58f423f8548fade0f

Issue #38, opened 2016-07-27
Menu items are blacked out on Ubuntu/Unity


takikawa:
DrRacket menu items are blacked out on Ubuntu/Unity (Xenial with Racket 6.6). Screenshot below:

![](http://i.imgur.com/TKOQPij.png)

They become un-blacked out when a new tab is opened. The tabs menu is also affected.

rfindler:
The first two menu items in the "view" tab are ones that set the label during on-demand, if that helps at all.

PR #112, opened 2018-08-11
Test wxme.rkt in Travis-CI (draft)

gus-massa:
This makes Travis-CI test wxme.rkt, but it has a few problems.

I had to use the minimal installer of Racket because the normal installer installs the default version of `gui-lib` and I wanted to use the version in the current branch. I tried using `raco pkg remove` and `raco pkg install` to replace it, but I get an error. I tried other combinations without success, so I finally decided to install the minimal version and then all the packages I need, but it is slower.

Another problem is that I can't install the `gui-test` package because it links to too many additional packages, so the installation is too slow and I even get a strange error. So ... I'm using the wxme.rkt file from the folder without installing the package. It's not a good idea but it works. In particular this is bad if we want later to test more files in Travis. 

I think that the problem is that `gui-test` uses then [sgl](https://github.com/racket/sgl) package, that builds the docs and then uses a lot of packages. I'm not sure if it's better to split `gui-test` in `gui-test-core` and `gui-test-more` or split the `sgl` package or just keep things as they are now. But I don't understand the details of neither of the packages to be sure.

---

A small detail is that I had to add `(exit 1)` in case wxme.rkt finds an error, because Travis uses the errorlevel to detect failures. With DrDr it's enough to write to the error port, but Travis doesn't care about that. Does this break something? (I also added a `flush-output` because in some test the final line was never printed.)

---

The other commit adds more test for `find-string`. They can be merged independently.

---

I'll push to this branch a  third commit soon. It adds some unnecessary messages to be sure that the test in Travis is using the test in _this_ branch with the code in _this_ branch, instead of the _default_ version of both. The idea is to not merge it, it's only for testing the tests. It also has a bad test, to be sure that Travis detects the error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+sudo: false
+
+language: c
+
+env:
+- PATH=~/racket/bin:$PATH
+
+before_install:
+- "export DISPLAY=:99.0"
+- "sh -e /etc/init.d/xvfb start"
+#- curl -L -o installer.sh http://plt.eecs.northwestern.edu/snapshots/current/installers/racket-test-current-x86_64-linux-precise.sh
+- curl -L -o installer.sh http://www.cs.utah.edu/plt/snapshots/current/installers/racket-current-x86_64-linux-precise.sh
+- sh installer.sh --in-place --dest ~/racket/
+
+install:
+- raco pkg install --auto -i --no-setup --skip-installed gui-test
+- racket -l- pkg/dirs-catalog --link --check-metadata pkgs-catalog .
+- echo file://`pwd`/pkgs-catalog/ > catalog-config.txt
+- raco pkg config catalogs >> catalog-config.txt
+- raco pkg config --set catalogs `cat catalog-config.txt`
+- raco pkg update -i --auto --no-setup gui-lib/ gui-doc/ gui/ tex-table/ gui-test/
+- raco setup --pkgs gui gui-lib gui-test tex-table
+- ls $HOME/.racket/download-cache
+
+script:
+- racket -l tests/gracket/wxme
+- raco test -e -l tests/gracket/test
+
+after_script:
--- a/gui-test/tests/gracket/wxme.rkt
+++ b/gui-test/tests/gracket/wxme.rkt
@@ -42,9 +42,12 @@
 
 (define (done)
   (printf "\n~a tests\n" test-cnt)
+  (flush-output)
   (if (zero? wrong-cnt)
       (printf "all passed\n")
-      (eprintf "~s FAILED\n" wrong-cnt)))
+      (begin
+        (eprintf "~s FAILED\n" wrong-cnt)
+        (exit 1))))
 
 ;; ----------------------------------------
 ;; String snips and lines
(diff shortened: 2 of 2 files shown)


gus-massa:
This repository doesn't have Travis enabled. In my copy:

[no flush #37](https://travis-ci.org/gus-massa/gui/builds/414848340) -> It's strange. I remember that it didn't show "_22366 tests - all passed_". Perhaps `flush-output` is not necessary, and the process in Travis continue to run an flush all the buffers eventually???

[with flush #38](https://travis-ci.org/gus-massa/gui/builds/414851277) -> Official version

[with error #39](https://travis-ci.org/gus-massa/gui/builds/414859441) -> This is the version that has an error on purpose.

samth:
I activated Travis on this repository. You might be interested in how the `typed-racket` repository installs packages to deal with the minimal-racket issue.

gus-massa:
I added a small commit to avoid some unnecessary output. It can be squashed with the new set of tests.

@samth I tried copying the travis.yml from `typed-racket` and `rackunit`,  with the obvious modifications of each one, and mixing them and other combinations, but I was not successful. I can retry later. (¿I guess I never used the line with `--skip-installed` [typed-racket/.travis.yml#L18](https://github.com/racket/typed-racket/blob/master/.travis.yml#L18)?)

samth:
That line is the one that adds `gui-test`, so it'll be important for you as well.

gus-massa:
I tried to copy the `install:` part of the `.travis.yml` file from `typed-racked` as close as possible. It's in another branch https://github.com/gus-massa/gui/commit/9ec329203659d99fbeb7afe67ac7b1534cc671b1 . But I get an [error](https://travis-ci.org/gus-massa/gui/jobs/414953839#L617) 

    collection-path: collection not found
      collection: "gui"
      in collection directories:
      [...]

samth:
The problem is that there's no collection named `gui`. It happens that the `typed-racket` package also has a `typed-racket` collection, so it works there.

Probably you went that line to be `raco setup --pkgs gui gui-lib gui-test tex-table`.

gus-massa:
@samth  Thanks. I think I fixed it.

gus-massa:
I changed the test to run with the the `-l` option, so it use the file from path of library instead .

I also read https://github.com/racket/gui/issues/82 and I added the checks that are in test.rktl. But I usually prefer to call each file individually. I'm using `-eQ` here, but I can use `--drdr` instead. Another option is to drop this part of the PR and add the other tests later.

gus-massa:
I think this can be merged now to be included in 7.1. I'm not sure if it's better to include the test in test.rkt or not. Any advice?

mflatt:
Loading the test from "test.rkt" seems like a good idea.

I don't have an objection to merging before the v7.1 branch, but unless it seems likely to uncover bugs during the release, I'd be conservative and just wait.

gus-massa:
I'll wait then. I don't expect to find something interesting, and I can try to configure my repository to run the test automatically just in case.

> Loading the test from "test.rkt" seems like a good idea.

I'm not sure if this mean that the current version is ok, or it means that it would be better to load each one of the four test that are loaded inside [test.rkt](https://github.com/racket/gui/blob/master/gui-test/tests/gracket/test.rkt) independently.

gus-massa:
I think it's better to call each of the test individually instead of using [gracket/test.rkt](https://github.com/racket/gui/blob/master/gui-test/tests/gracket/test.rkt) but they need some tweaking (for example some of them don't have `(require racket/gui)`), so I keep the simple version for now.

PR #116, opened 2018-10-07
Added alpha-channel to paste from clipboard

Kalimehtar:
Fixes https://github.com/racket/racket/issues/2273

### DIFF
--- a/gui-lib/mred/private/wx/win32/clipboard.rkt
+++ b/gui-lib/mred/private/wx/win32/clipboard.rkt
@@ -217,24 +217,25 @@
   (cond
    [(let ([bits (GetClipboardData CF_DIBV5)])
       (and bits
-        (begin0
-	  (and (= (BITMAPINFOHEADER-biBitCount (cast (GlobalLock bits) _pointer _BITMAPINFOHEADER-pointer)) 32) bits)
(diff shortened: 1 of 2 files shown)


Kalimehtar:
What is wrong with my propose? No merge, no comments. Why?

mflatt:
Apologies – I did lose track of this. I hope to take a look in the next couple of days and either merge (seems likely) or provide feedback.

mflatt:
Thank you very much for the patch!

I get a mangled image when I try to copy from Edge. It looks like Edge creates a 24-bit image (with no alpha channel) for `CF_DIBV5`, so parsing the data as a 32-bit image goes wrong.

One possibility is to pay attention to the `biBitCount` field when parsing the image. Another possibility is to use the `CF_DIBV5` data only when it is a 32-bit image, and otherwise use the other formats. I'd go with the second option, unless there's some advantage to `CF_DIBV5` besides an alpha channel. If that sounds like a good idea to you, can you make that change, and then I'll merge?

Kalimehtar:
> Another possibility is to use the CF_DIBV5 data only when it is a 32-bit image, and otherwise use the other formats.

I added the check.

PR #123, opened 2019-03-21
Start using the translation for font smoothing

alshopov:
Preferences -> Colors -> Racket and rest

Signed-off-by: Alexander Shopov <ash@kambanaria.org>


alshopov:
To be merged after https://github.com/racket/string-constants/pull/34

rfindler:
Oh, whoops: probably this also needs an update to info.rkt to depend on the latest string-constants pkg.

rfindler:
okay, I've done that part. Thanks very much for doing the rest!

PR #124, opened 2019-03-25
fix mismatched word-end-pos for autocomplete

yjqww6:
If the backspace key has been pressed during autocompletion, some following characters would be unexpectedly deleted when the selected string is inserted, because the widen-possible-completions method doesn't adjust the value of word-end-pos.

### DIFF
--- a/gui-lib/framework/private/text.rkt
+++ b/gui-lib/framework/private/text.rkt
@@ -3990,6 +3990,7 @@ designates the character that triggers autocompletion
         (let ([reasonable? (send completions-box widen)])
           (cond
             [reasonable?
+             (set! word-end-pos (sub1 word-end-pos))
              (let-values ([(_ __ x1p y1p) (send completions-box get-menu-coordinates)])
                (invalidate-bitmap-cache x0 y0 (max x1 x1p) (max y1 y1p)))]
             [else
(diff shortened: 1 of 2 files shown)


rfindler:
Here is a test case:

```
#lang racket
(require framework racket/gui/base rackunit)
(define t (new (class (text:autocomplete-mixin text%)
                 (define/override (get-all-words)
                   '("abcd"))
                 (super-new))))
(define f (new frame% [label ""]))
(define ec (new editor-canvas% [parent f] [editor t]))

(send t insert "pqr abc xyz")
(send t set-position 7 7)
(send t auto-complete)
(send ec on-char (new key-event%
                      [key-code #\backspace]))
(send ec on-char (new key-event%
                      [key-code #\return]))

(check-equal? (send t get-text) "pqr abcd xyz")
```

I think this could go in tests/framework/text. Or maybe it makes sense to start a new file with autocomplete tests, as they seem to be quite lacking!

rfindler:
Thanks! This looks good to me. Shall I push to master?

rfindler:
Oh, actually, it shoudl be one commit. But I can do that if you want.

yjqww6:
Commits squashed, I think it is ready.

rfindler:
Thanks!

PR #125, opened 2019-03-25
make comparator for autocompletion candidates asymmetric

yjqww6:
The original comparator is not asymmetric, which leads to confusing results. 

For example, when I enter "define-syntax-", "syntax" is usually more preferred than "define-syntax-rule" on my machine.

The old one:
```
> (define comp
    (match-lambda**
     [((cons w r) (cons w* r*))
      (or (> r r*)
          ;; prefer shorter matches
          (< (string-length w) (string-length w*)))]))
> (comp '("define-syntax-rule" . +inf.0) '("syntax" . 2))
#t
> (comp '("syntax" . 2) '("define-syntax-rule" . +inf.0))
#t
```

### DIFF
--- a/gui-lib/framework/private/autocomplete.rkt
+++ b/gui-lib/framework/private/autocomplete.rkt
@@ -3,6 +3,9 @@
 (require racket/class racket/match)
 (provide autocompletion-cursor<%> autocompletion-cursor%)
 
+(module+ test
+  (require rackunit))
+
 (define autocompletion-cursor<%>
   (interface ()
     get-completions  ;      -> (listof string) 
@@ -38,6 +41,14 @@
                             [else c]))]))
           (length parts)))
 
+(define better-completion?
+  (match-lambda**
+   [((cons w r) (cons w* r*))
+    (cond
+      [(not (= r r*)) (> r r*)]
+      ;; prefer shorter matches
+      [else (< (string-length w) (string-length w*))])]))
+
 ;; ============================================================
 ;; autocompletion-cursor<%> implementation
 
@@ -65,10 +76,7 @@
                             [r (in-value (rnk w))]
                             #:when (>= r mx))
                   (cons w r))
-                (match-lambda** [((cons w r) (cons w* r*))
-                                 (or (> r r*)
-                                     ;; prefer shorter matches
-                                     (< (string-length w) (string-length w*)))]))))
+                better-completion?)))
     
     (define all-completions-length (length all-completions))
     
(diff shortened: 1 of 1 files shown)


rfindler:
Can you help with a test case, too? Here is a (bad) attempt because it doesn't distinguish the code before and after your commit (the expected results are just bogus for now):

```
(module+ test
  (check-equal? (send (new autocompletion-cursor%
                           [word "x"]
                           [all-words '("define-syntax-rule"
                                        "syntax")])
                      get-completions)
                1)
  (check-equal? (send (new autocompletion-cursor%
                           [word "x"]
                           [all-words '("syntax"
                                        "define-syntax-rule")])
                      get-completions)
                2))
```

You would also need to add this line (preferably near the top of the file):
```
(module+ test (require rackunit))
```

yjqww6:
@rfindler I think the following test asserts a stable order that distinguishes them. If you think it is ok, I will add it as gui/gui-test/framework/tests/private/autocomplete.rkt(I think this is the place, right?)
```
#lang racket
(require framework/private/autocomplete)

(define (get-completions word ls)
  (send (new autocompletion-cursor%
             [word word]
             [all-words ls])
        get-completions))

(module+ test
  (require rackunit)

  (define word "define-syntax-")
  (define candidates '("define-syntax-rule" "define-syntax" "syntax"))

  (match-define (and (list a b c) expected)
    (get-completions word candidates))

  (check-equal? (get-completions word (list a c b))
                expected)

  (check-equal? (get-completions word (list b a c))
                expected)

  (check-equal? (get-completions word (list b c a))
                expected)

  (check-equal? (get-completions word (list c a b))
                expected)

  (check-equal? (get-completions word (list c b a))
                expected)
  
  )
```

rfindler:
Yes, putting the test there works. If you did start a new file for autocomplete tests, you could put it there too, tho. As you wish.

It might be better to move the comparison predicate into its own function, tho, so you don't have to work so hard and rely on how `sort` works for an effective test, tho.

yjqww6:
@rfindler Sorry for the edits, I missed your points at first, I found some submodule tests exist inside the gui-lib directory then I understand you mean appending it to the same file.

yjqww6:
@rfindler Or if you think moving into gui-test is better, I will do the changes.

rfindler:
I'm happy either way: with a `test` submodule in that one file or adding them to the `gui-test` pkg. Whatever you think is best is good with me.

yjqww6:
Thanks, I leave it as a submodule in that.

rfindler:
What do you think about moving the predicate function to the top-level of the file, tho, so that it can be tested directly?

yjqww6:
That should be good, I apply the changes.

rfindler:
Thanks!

PR #128, opened 2019-04-22
tex-table: more logical operators

AlexKnauth:
This adds `\lnot`, `\land`, and `\lor` for propositional logic, `\lozenge` and `\square` for modal logic (`\lozenge` is used for Pollen as well), both kinds of ampersands for linear logic, and the lollipop connectives for linear logic and computability logic.

### DIFF
--- a/tex-table/tex-table.rkt
+++ b/tex-table/tex-table.rkt
@@ -109,6 +109,8 @@
     ("pm" "±")
     ("cap" "∩")
     ("diamond" "◇")
+    ("lozenge" "◊")
+    ("square" "□")
     ("oplus" "⊕")
     ("mp" "∓")
     ("cup" "∪")
@@ -126,6 +128,9 @@
     ("sqcup" "⊔")
     ("vee" "∨")
     ("wedge" "∧")
+    ("lor" "∨")
+    ("land" "∧")
+    ("lnot" "¬")
     ("triangleleft" "◃")
     ("odot" "⊙")
     ("star" "★")
@@ -201,7 +206,17 @@
     ("ldots" "…")
     
     ("langle" "⟨")
-    ("rangle" "⟩")))
+    ("rangle" "⟩")
+
+    ("amp" "&")
+    ("invamp" "⅋")
+
+    ("multimap" "⊸")
+    ("leftmultimap" "⟜")
+    ("rightlollipop" "⊸")
+    ("leftlollipop" "⟜")
+    ("uplollipop" "⫯")
+    ("downlollipop" "⫰")))
 
 (module+ test
   (require racket/match)


rfindler:
Several of these are already in the file, with their names matching their latex counterparts.

AlexKnauth:
In LaTeX, `\land`, `\lor`, and `\lnot` exist as aliases for `\wedge`, `\vee` and `\neg`, but its much more convenient to to use `\land`, `\lor`, and `\lnot` since they are much more consistently named.

rfindler:
Ah, I didn't know that.

Does the DrRacket documentation build with this commit in place?

AlexKnauth:
Yes. Although I think I might want to change `\turnedampersand` to `\invamp` to follow the LaTeX `cmll` package, and I might want to re-order them so that "lozange" "square" and "diamond" all appear together, and the "land", "lor", and "lnot" appear next to the "wedge", "vee", and "neg" entries in the documentation.

AlexKnauth:
And here's how the rendered docs look:
image

rfindler:
Those sound like nice changes to me.

And I meant the PDF docs, sorry.

Robby

rfindler:
I can look into the PDF changes if you want. Usually changing this file
breaks it. :)

Robby

AlexKnauth:
Oh. How do I build the PDF docs?

rfindler:
Raco setup has an option. Scribble on the main DrRacket manual file also
works.

AlexKnauth:
Running scribble in PDF mode on `drracket.scrbl` does NOT succeed. I'm getting this error message:
```
run-pdflatex: got error exit code
  context...:
   .../scribble-lib/scribble/private/run-pdflatex.rkt:34:2: loop
   .../scribble-lib/scribble/private/run-pdflatex.rkt:16:0: run
   .../scribble-lib/scribble/private/indirect-renderer.rkt:38:8: for-loop
   .../more-scheme.rkt:261:28
   .../scribble-lib/scribble/render.rkt:33:0: render57
   .../racket/contract/private/arrow-val-first.rkt:430:3
   ".../scribble-lib/scribble/run.rkt": [running body]
   temp37_0
   for-loop
   run-module-instance!125
   ".../raco/raco.rkt": [running body]
   temp37_0
   for-loop
   run-module-instance!125
   ".../raco/main.rkt": [running body]
   temp37_0
   ...
```

rfindler:
Right, that's the expected error. It isn't very informative because you aren't getting the latex-level error. The latex level error will lead to the fix. Probably I should do this part. I'll try to get to it before too long.

rfindler:
Okay, it looks like \invamp, the multimap and the lollipop symbols are all not in the latex libraries that DrRacket is already importing (at least not with those names). I see that they are in AMS symbols but I think something goes wrong when you import that. If you want to try to figure this out, run 

scribble --latex --dest x drracket.scrbl

and then fiddle with the latex until it builds and then try to port things back. square looks easy (just requires updating scribble/latex-render (search for `Diamond` for the big table) but the others that don't work look like they'll require changing the libraries that the latex code uses which can get a little complicated if the libraries conflict with each other.

rfindler:
I'm not sure of the best choice, but looking at [symbols-letter.pdf](http://tug.ctan.org/info/symbols/comprehensive/symbols-letter.pdf), it seems like the txfonts package both works with DrRacket's current pile of latex and contains the symbols you want to add.

Do you think that's the best choice or would one of the other multimap-including latex packages be better?

rfindler:
@AlexKnauth how about we just remove the (few?) offending latex symbols for now and deal with them separately, later?

AlexKnauth:
Okay, I will be able to get to that next week.

rfindler:
I'm happy to do it if you're happy with the plan?

AlexKnauth:
If you'd be happy to, I'm fine with you pushing to this branch, or including any code written here in a different branch, whichever you prefer.

rfindler:
Okay, I've pushed something for these. I couldn't find the up and down lolli so I just got rid of those.

Meanwhile, there are many many more symbols that might have unicode equivalents that can now be added if you are interested. The new symbols are in [this pdf document](http://mirror.las.iastate.edu/tex-archive/fonts/txfonts/doc/txfontsdoc.pdf) and adding them to tex-table.rkt requires additions to scribble/latex-render.rkt along the lines of [this commit](https://github.com/racket/scribble/commit/c62d008cfb4cbefe59b29ed245e73c82eafb08b0).

I'll leave this open for now in case someone wants to add more stuff as it has a kind of roadmap.

rfindler:
Looks like no one is going to work more on this, closing.

PR #134, opened 2019-06-17
horizontal-/vertical-panel% make style parameters consistent with panel%

SimonLSchlee:
Documentation: adds the 'hide-hscroll and 'hide-vscroll style parameters from panel% to vertical-panel% and horizontal-panel%


mflatt:
Thanks!

PR #140, opened 2019-08-06
Fix xvfb in travis-ci

gus-massa:
There was some change in Travis-ci, and the test are failing. This is the new recomended method to start xvfb.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ language: c
 env:
 - PATH=~/racket/bin:$PATH
 
+services:
+- xvfb
+
 before_install:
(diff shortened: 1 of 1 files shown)


gus-massa:
What about merging this? The test are still failing because `xvfb` is not working.

samth:
Sorry, this dropped off my radar.

gus-massa:
There are also:

- https://github.com/racket/compatibility/pull/7 But it has some weird interaction with the order packages are installed ...

- https://github.com/racket/r6rs/pull/4 But it need some editorial decision about what is an error and it also need some modifications to make it travis-friendly ...

So they need more work before merging.

Issue #144, opened 2019-10-02
list-box% callbacks improperly invoked on MacOS


jgreco:
Neither `list-box%`'s `select` nor `list-control<%>`'s `set-selection` are meant to fire a `list-box%`'s callback.   However on MacOS, they both do.

## Test Code

```racket
#lang racket/gui

(define f (new frame% [label "lb test"]))
(define lb (new list-box%
                [label "listbox"]
                [choices '("one" "two" "three")]
                [parent f]
                [callback (lambda (lb e)
                            (printf "selected: ~a~n" (send lb get-string-selection)))]))
(send f show #t)

;; per: https://docs.racket-lang.org/gui/list-box_.html
;;  "The control’s callback procedure is not invoked."
(send lb select 0)

;; per: https://docs.racket-lang.org/gui/list-control___.html
;;  "The control’s callback procedure is not invoked when this method is called."
(send lb set-selection 1)
(send lb set-selection 2)
```

### Expected Results:
[nothing]

### Results on MacOS:
```
selected: three
selected: three
selected: three
```

If a `(sleep/yield 1)` expression is placed in between the selects, then the output (still incorrect) changes to:
```
selected: one
selected: two
selected: three
```

### Results on Linux:
[nothing]

jgreco:
(Fixed with f0e4cdb : https://github.com/racket/gui/pull/146#issuecomment-565731108)

PR #146, opened 2019-11-04
fix list-box% selection callback on MacOS

jgreco:
Addresses issue #144

(According to the docs, the select and set-selection methods of list-box% should not invoke the list-box% callback.)

Tested with:
```racket
#lang racket/gui
;;; This program should print nothing when everything is working.

(define f (new frame% [label "lb test"]))
(define lb (new list-box%
                [label "listbox"]
                [choices '("one" "two" "three")]
                [parent f]
                [callback (lambda (lb e)
                            (printf "lb selected: ~a~n" (send lb get-string-selection)))]))
(define ch (new choice%
                [label "choices"]
                [choices '("one" "two" "three")]
                [parent f]
                [callback (lambda (lb e)
                            (printf "ch selected: ~a~n" (send lb get-string-selection)))]))
(define mlb (new list-box%
                [label "listbox"]
                [choices '("one" "two" "three")]
                [style '(multiple)]
                [parent f]
                [callback (lambda (lb e)
                            (printf "lb selected: ~a~n" (send lb get-string-selection)))]))
(send f show #t)

;; per: https://docs.racket-lang.org/gui/list-box_.html
;;  "The control’s callback procedure is not invoked."
(send lb select 0)

;; per: https://docs.racket-lang.org/gui/list-control___.html
;;  "The control’s callback procedure is not invoked when this method is called."
(send lb set-selection 1)
(send lb set-selection 2)

;; multi-selection list-box%
(send mlb select 0 )
(send mlb select 2 )
(send mlb select 1 )
(send mlb select 1 #f )

;; choice% is implemented differently so it should be fine
(send ch set-selection 0)
(send ch set-selection 1)

```

### DIFF
--- a/gui-lib/mred/private/wx/cocoa/list-box.rkt
+++ b/gui-lib/mred/private/wx/cocoa/list-box.rkt
@@ -2,7 +2,7 @@
 (require ffi/unsafe/objc
          ffi/unsafe
          racket/class
-         (only-in racket/list take drop)
+         (only-in racket/list index-of take drop)
           "../../syntax.rkt"
           "../../lock.rkt"
          "item.rkt"
@@ -86,6 +86,8 @@
   (define num-columns (length columns))
   (define data (map (lambda (x) (box #f)) choices))
   (define count (length choices))
+  (define ignore-selections '())
+  (define last-selections '())
 
   (define cocoa (as-objc-allocation
                  (tell (tell NSScrollView alloc) init)))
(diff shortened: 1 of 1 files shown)


mflatt:
Thank you for the report and patch! I've pushed a different change that I think more directly addresses the problem. If you have a chance to look at commit f0e4cdbbbd91159ca78933cd51d19ce209207b24, I'd appreciate feedback on whether it makes sense and works for you.

jgreco:
That works great, thanks mflatt!

Issue #119, opened 2018-11-27
contract violation when comparing snip% objects using equal?


alex-hhh:
When the following program is run in DrRacket 7.1:

```racket
#lang racket
(require racket/gui pict pict/snip)
(define snip1 (make-object image-snip% (pict->bitmap (text "hello"))))
(define snip2 (make-object pict-snip% (text "hello")))
```

Trying to compare `snip1` and `snip2` using `equal?` I get a contract violation when trying to compare the two snips, but only if the `pict-snip%` is the first argument:

```
> (equal? snip1 snip2)
#f
> (equal? snip2 snip1)
. . other-equal-to?: contract violation
  expected: (is-a?/c image-snip%)
  given: (object:pict-snip% ...)
  in: the 1st argument of
      the other-equal-to? method in
      image-snip%/c
  contract from: 
      <pkgs>/snip-lib/racket/snip.rkt
  contract on: image-snip%
  blaming: E:\Alex\t\snip.rkt
   (assuming the contract is correct)
  at: <pkgs>/snip-lib/racket/snip.rkt:26.19
> 
```

rfindler:
This appears to be a check in `image-snip%` that is intentional, but I'm not sure it should be.

samth:
I think "equal? doesn't error" is an invariant we expect to hold, and therefore this check is wrong.

alex-hhh:
I am not sure that the contract on the `other-equal-to?` method is wrong.  The `equal?` mechanism works fine if the `image-snip%` is compared against a string, for example:

```
> (equal? snip1 "hello")
#f
> (equal? "hello" snip1)
#f
> 
```

I suspect that the code that calls `other-equal-to?` is incorrect, but I am not familiar enough with the implementation.

rmculpepper:
@alex-hhh `equal?` only calls user-defined implementations for structs (specified via `prop:equal+hash-code`) if the two struct instances have the same extension. That is, they have a common struct type with the `prop:equal+hash-code` property, and neither of them have a more specific struct type that redeclares that property. For classes, that translates to classes declaring that they implement the `equal<%>` interface. The bug is that the `snip%` class declares it implements the `equal<%>` interface and `image-snip%` just extends `snip%`, so its `other-equal-to` method should only assume it gets a `snip%`, not specifically a `image-snip%`.

I think just changing `image-snip%` to list `equal<%>` as an interface should fix this bug.

@samth `equal?` can error because it forces a delayed contract:

[...]

PR #148, opened 2019-11-23
Update contract on image-snip% other-equal-to? method (racket/gui#119)

alex-hhh:
Update the signature of the `other-equal-to?` method for `image-snip%`, since
it now accepts a `snip%` as an argument.  Also added test cases for checking
that `image-snip%` objects can be compared with other objects.

This commit corresponds to the fix for issue #119, which is in the racket/snip
package.


alex-hhh:
This pull request is additional to https://github.com/racket/snip/pull/6, I added a test case, but it will fail unless the snip library is updated with the changes referenced in that pull request.

alex-hhh:
This test fails, as expected, since the fix in https://github.com/racket/snip/pull/6 has not been merged.

PR #150, opened 2019-12-13
pass the correct value to set-wheel-steps

Bogdanp:
I was looking at the scrolling behavior changes in db864a46597ee618ba55363bbf8bf4b6ebda90a7 and my little linter tool picked up this problem.


mflatt:
Thanks!

PR #151, opened 2019-12-14
improve editor-canvas% scrolling responsiveness

Bogdanp:
This significantly improves trackpad scrolling responsiveness for me on macOS. I don't have a mouse to test with or any Windows or Linux machines so it would be helpful if folks with access to any of those things could give this a spin.  Steps to test:

1. clone this repo and checkout this commit
2. run `raco pkg update --link gui-lib` from the repository root
3. run `racket -l drracket` and open a large file
4. scroll around

Scrolling is still a little laggy when there are multiple tabs open. I'm not sure why that is yet.

### DIFF
--- a/gui-lib/mred/private/wxme/editor-canvas.rkt
+++ b/gui-lib/mred/private/wxme/editor-canvas.rkt
@@ -8,7 +8,7 @@
          racket/snip/private/prefs
          racket/snip/private/private
          (only-in "cycle.rkt" popup-menu%)
-         (only-in "../kernel.rkt" queue-refresh-event)
+         (only-in "../kernel.rkt" queue-callback queue-refresh-event)
          "wx.rkt")
 
 (provide editor-canvas%)
@@ -142,8 +142,8 @@
 (define default-wheel-amt
   (let ([v (get-preference* 'GRacket:wheelStep)])
     (if (exact-integer? v)
-        (max 3 (min 1000 v))
-        3)))
+        (max 1 (min 1000 v))
+        1)))
 
 (define (INIT-SB style)
   (append
@@ -282,6 +282,10 @@
   (define scroll-width (if fake-x-scroll? 1 1)) ;; else used to be 0
   (define scroll-height (if fake-y-scroll? 1 1))
 
+  (define scroll-events null)
+  (define scroll-flush-scheduled? #f)
+  (define scroll-delay 16)
+
   (define hscrolls-per-page 1)
   (define vscrolls-per-page 1)
   (define hpixels-per-scroll 0)
@@ -305,7 +309,7 @@
       (set! blink-timer #f))
     (send admin set-canvas #f)
     #|(super ~)|#)
-  
+
   (define/override (on-size)
     (unless noloop?
       (queue-refresh-event
@@ -442,14 +446,14 @@
                    (send media adjust-cursor event))))
            (when media
              (send media on-event event))))
-        
+
         (when (send event dragging?)
           (when out-of-client?
             ;; Dragging outside the canvas: auto-generate more events because the buffer
             ;; is probably scrolling. But make sure we're shown.
             (when (is-shown-to-root?)
               (set! auto-dragger (parameterize ([current-eventspace (get-eventspace)])
-                                   (new auto-drag-timer% 
+                                   (new auto-drag-timer%
                                         [canvas this]
                                         [event event])))))))))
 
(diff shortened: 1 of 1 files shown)


gui-lib/mred/private/wxme/editor-canvas.rkt
```diff
@@ -466,49 +500,78 @@
...
+      (do-scroll x y #t ox oy))
+
+    (collect-garbage 'incremental))
+
+  ;; Called by the scroll-timer whenever scrolling events need to be
+  ;; applied.  Roughly once every 33ms if the canvas is focused.
+  (define/public (on-scroll-flush)
+    (unless (null? scroll-events)
+      (process-scroll-events (reverse scroll-events))
+      (set! scroll-events null)))
```
mflatt: Probably better to reverse those two lines, just in case `do-scroll` somehow leads to a `yield`.

gui-lib/mred/private/wxme/editor-canvas.rkt
```diff
@@ -466,49 +500,78 @@
...
+                (or new-y y))))
+
+    (when (and x y)
+      (do-scroll x y #t ox oy))
+
+    (collect-garbage 'incremental))
+
+  ;; Called by the scroll-timer whenever scrolling events need to be
+  ;; applied.  Roughly once every 33ms if the canvas is focused.
+  (define/public (on-scroll-flush)
```
mflatt: Maybe check that the canvas is still shown, just in case, like the check in `blink-timer%`?

mflatt:
Aside from my two small comments, this looks generally good and works for me.

Bogdanp:
Thanks @mflatt! I've made the changes you mentioned. 

Since @rfindler mentioned this isn't an improvement on his machine, I'd like to hold off on merging this change until my own 2019 MBP arrives so I can test things in similar conditions to his and see if I can make this behave better on newer/more powerful machines.

rfindler:
In a slack conversation, @Bogdanp asked if reversing the direction of the scroll worked better with this pull request or on origin/master and it definitely works better with the pull request. Just scrolling in one direction I can't really tell a difference between 7a4f2885611e19dece59b93ffc335c5bb06b0129 and the version on origin/master.

gui-lib/mred/private/wxme/editor-canvas.rkt
```diff
@@ -1066,20 +1122,26 @@
 
   (define/override (set-scrollbars x y x2 y2 x3 y3 x4 y4 ?) (void))
 
-  (define/public (get-scroll x y)
+  (define/public (get-scroll x-out y-out)
+    (define-values (x y) (get-scroll/values))
+    (set-box! x-out x)
+    (set-box! y-out y))
+
+  (define/public (get-scroll/values)
```
rfindler: Should this be a private method?

gui-lib/mred/private/wxme/editor-canvas.rkt
```diff
@@ -466,49 +515,56 @@
 
   (define/public (popup-for-editor b m) #f)
 
+  ;; Reduces an ordered set of scroll events into the next scroll
+  ;; position and then scrolls to it.
+  (define (process-scroll-events events)
```
rfindler: I'm not completely sure, but this looks like a field bound to a lambda? Maybe it should be a private method instead.

gui-lib/mred/private/wxme/editor-canvas.rkt
```diff
@@ -139,6 +139,44 @@
...
+    (define (loop)
+      (when running
+        (define deadline
+          (alarm-evt (+ (current-inexact-milliseconds) interval)))
+        (when canvas
+          (if (send canvas is-shown-to-root?)
+              (send canvas on-scroll-flush)
+              (stop)))
+        (yield deadline)
+        (loop)))
```
mflatt: A loop that `yield`s does not seem like a good idea, because it inverts control in a way that is less composable. For example, imagine with two such loops; the second one cannot return and continue until the inner one ends. I guess the idea here is that the loops would be limited by input events. Still, I think using a `timer%` or a looping thread with `queue-callback` (which should differ only in terms of priority) is the way to go.

Bogdanp: Thanks! That makes sense. I've reverted back to the `timer%` implementation.

gui-lib/mred/private/wxme/editor-canvas.rkt
Bogdanp: I went ahead and merged `process-scroll-events` into `on-scroll-flush`. There wasn't much point in having them be separate.

Bogdanp:
I've picked up a mouse and tested this on macOS and on Ubuntu (inside a VM) with both the trackpad and the mouse and everything looks good. I think this is good to go.

rfindler:
It looks like `on-scroll-flush` is a new public method. Is that intentional? If so, it should be documented; if not, then probably there needs to be a `define-local-member-name` in some appropriate scope.

Otherwise, looks good to me.

mflatt:
This is at the `wx` level, so "public" is not really public.

rfindler:
Ah, okay.

One other question I thought of: what happened with the concern about machines with different performance that required setting `SCROLL-DELAY` to different values. What was the resolution with that?

Bogdanp:
> One other question I thought of: what happened with the concern about machines with different performance that required setting SCROLL-DELAY to different values. What was the resolution with that?

I also tested this on an external 4k monitor that refreshes at 30hz and having the scroll timer pinned to ~60 FPS works great. I think the worst case for slow machines is that scrolling will be a little choppy, but it should still be an improvement because those machines probably suffer from the most lag (my thinking is that their input devices likely deliver events more frequently than 60 FPS, so the scroll-handling code currently has to run more frequently and that means scrolling is more laggy). I can't say for sure, though, because I don't have any machines that slow. Personally, I'd prefer choppy-but-responsive scrolling over laggy scrolling.

mflatt:
@Bogdanp It looks like the scroll timer is always running, even when there are no scroll events. That makes even plain `GRacket` take 5% of a CPU even when sitting in the background. Am I understanding correctly? Is there a problem with enabling the timer only when scroll events are pending?

Bogdanp:
D'oh.  Thanks, @mflatt!  What do you think about the approach in 7fb1c9e?  Instead of using a `timer%`, I've switched to scheduling flushes on demand because starting the timer on demand made things choppy.  This avoids using up unnecessary CPU and I don't think it has the same problems that looping with a yield did.  Scrolling with this change in place also feels smoother on my machine.

mflatt:
I think we really want to avoid `yield`. Instead of `(queue-callback (lambda () (yield alarm) ....)))`, does something like `(thread (lambda () (sleep timeout) (queue-callback (lambda () ....)))))` work smoothly?

Bogdanp:
Yup! That works great, too.

mflatt:
Thanks! Can you squash and I'll take one more look from the top (and then likely merge)?

Bogdanp:
@mflatt done.

PR #152, opened 2019-12-14
bubble up unhandled scroll events

Bogdanp:
I noticed that `NSScrollView`-based widgets stopped scrolling on master since db864a46597ee618ba55363bbf8bf4b6ebda90a7. I still haven't fully wrapped my head around `do-key-event` so feel free to scrap this if there's a better approach.

You can use this script to reproduce the issue:

```racket
#lang racket/base

(require racket/gui)

(define frame
  (new frame%
       [label "Test"]
       [width 500]
       [height 500]))

(define lb
  (new list-box%
       [parent frame]
       [label #f]
       [choices (make-list 120 "a")]))

(send frame show #t)
```

### DIFF
--- a/gui-lib/mred/private/wx/cocoa/window.rkt
+++ b/gui-lib/mred/private/wx/cocoa/window.rkt
@@ -456,8 +456,10 @@
   (cond
     [(not wx) #f]
     [else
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #153, opened 2019-12-20
adjust the style for MMTabBarView on Mojave and later

rfindler:


rfindler:
I think this probably needs to update a dependency version number for the actual library too.

mflatt:
Merged with the version update, now that the "gui-x86_64-macosx" update is available through the package server.

PR #161, opened 2020-01-24
add ctl-shift-z as redo in windows

jbclements:
This pull request is kind of just spit-balling; it looks to me like it should work, and it compiles on my mac, but

1) I haven't actually tried it on a Windows system (could maybe try this tonight), but more importantly,

2) I see that d:s:z works on macosx, but I can't see where that's specified in the source code; presumably, adding the c:s:z binding in Windows should be done in the same way that adding d:s:z is done for mac.

Can anyone point me to the place in the code where d:s:z is set up for the mac?

### DIFF
--- a/gui-lib/framework/private/keymap.rkt
+++ b/gui-lib/framework/private/keymap.rkt
@@ -572,6 +572,8 @@
                     key)
                    func))])
       (add/map "editor-undo" 'undo "z")
+      (when (eq? (system-type) 'windows)
+        (add/map "editor-redo" 'redo "s:z"))
       (unless (eq? (system-type) 'macosx)
         (add/map "editor-redo" 'redo "y"))
       (add/map "editor-cut" 'cut "x")


jbclements:
Relevant to #161

rfindler:
I think you'd have to make a new item around here:

  https://github.com/racket/gui/blob/master/gui-lib/framework/private/standard-menus-items.rkt#L295

Probably you'll want to change a number of things as we don't want there to be two different callbacks for the two different menu items.

And I'm not really two menu items is a good idea. Is there really precedent for this?

jbclements:
I'm not sure either. Keavon Chambers (a student in my PL class) points to 

https://github.com/notepad-plus-plus/notepad-plus-plus/issues/856

a discussion from 2015 of how every Windows editor on earth supports c:s:z. Since that writing, it appears that Microsoft Word has joined this list. Also, it does seem to increase mac-windows uniformity.

rfindler:
I didn't see a discussion of what actually shows up in the menus of these apps. I think we can certainly change the shortcut in the menu and we can certainly add a keybinding somewhere.

Looking at the code, I'm not sure that `keymap:get-editor` is used very much (which might be its own set of bugs).

PR #162, opened 2020-02-09
adds auto-delete for some characters created by auto-parens

bzyeung:
Addresses racket/drracket#201 regarding parens as well as analogous
behavior for " and | characters.

### DIFF
--- a/gui-doc/scribblings/framework/color.scrbl
+++ b/gui-doc/scribblings/framework/color.scrbl
@@ -270,6 +270,15 @@
     editor.
 
   }
+  @defmethod[(get-matching-paren-string [paren-str string?]
+                                        [get-side (or/c 'open 'close 'either) 'either])
+             (or/c string? #f)]{
+    Returns a string of a paren matching the other side of @racket[paren-str] as specified
+    by the @racket[pairs] argument of @method[color:text<%> start-colorer], if one exists
+    on the side indicated by @racket[get-side]. If there is no match on the corresponding
+    side, including if @racket[paren-str] contains any characters other than a single paren
+    token (even whitespace), returns @racket[#f] instead.
+  }
   @defmethod[(skip-whitespace (position exact-nonnegative-integer?)
                               (direction (or/c 'forward 'backward))
                               (comments? boolean?))
--- a/gui-lib/framework/private/color.rkt
+++ b/gui-lib/framework/private/color.rkt
@@ -66,6 +66,8 @@ added get-regions
     
     is-lexer-valid?
     on-lexer-valid
+
+    get-matching-paren-string
     
     skip-whitespace
     backward-match
@@ -743,6 +745,21 @@ added get-regions
     (define/private (get-match-color) 
       (color-prefs:lookup-in-color-scheme 'framework:paren-match-color))
     
+    ;; See docs
+    ;; String Symbol -> (U String False)
+    (define/public (get-matching-paren-string paren-str [get-side 'either])
+      (define sym (string->symbol paren-str))
+      (define checker
+        (case get-side
+          [(either) (λ (x) (member sym x))]
+          [(open) (λ (x) (equal? sym (cadr x)))]
+          [(close) (λ (x) (equal? sym (car x)))]
+          [else (raise-argument-error 'get-matching-paren-string
+                                      "(or/c 'open 'close 'either)"
+                                      get-side)]))
+      (for/first ([pair (in-list pairs)]
+                  #:when (checker pair))
+        (symbol->string (car (remove sym pair)))))
     
     ;; higlight : number number number (or/c color any)
     ;;   if color is a color, then it uses that color to higlight
--- a/gui-test/framework/tests/racket.rkt
+++ b/gui-test/framework/tests/racket.rkt
@@ -8,12 +8,35 @@
 
 (module+ test
   (with-private-prefs
+    (test-get-matching-paren-string)
     (open-paren-typing)
     (test-text-balanced)
     (indentation-tests)
     (magic-square-bracket-tests)
     (insert-return-tests)
     (test-message-send)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;; testing get-matching-paren-string method
+;;
+
+(define (test-get-matching-paren-string)
+  (define t (new racket:text%))
+
+  (check-equal? (send t get-matching-paren-string "(") ")")
+  (check-equal? (send t get-matching-paren-string "(" 'close) ")")
+  (check-equal? (send t get-matching-paren-string "(" 'open) #f)
+  (check-equal? (send t get-matching-paren-string "]") "[")
+  (check-equal? (send t get-matching-paren-string "]" 'open) "[")
+  (check-equal? (send t get-matching-paren-string "{" 'either) "}")
+  (check-equal? (send t get-matching-paren-string "}" 'close) #f)
+  (check-exn exn:fail? (λ () (send t get-matching-paren-string "(" #f)))
+  (check-exn exn:fail? (λ () (send t get-matching-paren-string "(" 'forward)))
+  (check-equal? (send t get-matching-paren-string "[]") #f)
+  (check-equal? (send t get-matching-paren-string "} ") #f)
+  (check-equal? (send t get-matching-paren-string "") #f)
+  (check-equal? (send t get-matching-paren-string "abc") #f))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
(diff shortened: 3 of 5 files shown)


rfindler:
I didn't try this out but just reading the code it looks pretty reasonable to me.

One question I wonder about: should we be getting the matching pairs from the colorer instead of having them be directly in the text? I see that this is happening for the paren matching pairs but not the double quotes or bars.

bzyeung:
As far as I'm aware, it's not getting the paren pairs from the colorer, either.  `racket-paren:get-paren-pairs` is just a list of corresponding pairs of single-character strings.  I noticed that the list was used in a couple of other places throughout the file, so I used it here, too.

bzyeung:
To be a bit more precise, I was using the colorer to determine if the two characters were the same classification.  That might be more conservative than necessary for non-parens, but I was trying to err towards conservative behavior when there was potential ambiguity.

bzyeung:
Looking at it again, it also has that `eq?` instead of `equal?` issue that you had mentioned in the other bugfix.  I'm happy to tweak that along with any other recommendations you have.

rfindler:
Maybe the right thing is to add a method to [`color:text<%>`](https://github.com/racket/gui/blob/master/gui-lib/framework/private/color.rkt#L52) (which also means we need docs for it) that accepts something that might be one half of a paren pair (the things in the [`pair` private field](https://github.com/racket/gui/blob/master/gui-lib/framework/private/color.rkt#L120)) and, if it is to return the other side of the pair and if it isn't to return `#f`; then we could use that here.

rfindler:
Let me know if you have questions about any of that or want to hand things off.

bzyeung:
Sorry about going silent for a bit there!  I think I follow that, and it makes sense to me.  Does that mean that (at least in theory) we would eventually want to replace all uses of `racket-paren:get-paren-pairs`?  That is, that those uses should be handled at the `color:text<%>` level with appropriate manipulations of the `pair` field?

rfindler:
I suppose it depends on why they are using them! But possibly, yes.

bzyeung:
As I started exploring this, I thought back to your initial musing about potentially asymmetrical treatment of parens and non-paren paired characters.  If we go forward with the change to the colorer, wouldn't this *make* it asymmetrical?

Currently, the [`pair` private field](https://github.com/racket/gui/blob/master/gui-lib/framework/private/color.rkt#L120) only references parens.  Wouldn't extending it to double quotes and bars be problematic?  (as in the [`insert-close-paren` method](https://github.com/racket/gui/blob/547211f6dbe1400b2c89dcb8ef1eaaa52239b7f8/gui-lib/framework/private/color.rkt#L1101)) Maybe the right answer is to add another field for string-delimiters and symbol-delimiters (for example), with associated methods, but that's starting to seem like a rather large change.  Either way, I'm wondering whether this should go in a future commit instead of this one.  (And yes, I'm volunteering to work on that, if you decide that's the way to go!  :) )

For sure, sticking entirely to text patterns, especially forcing the two positions to be the same token type, was a simplification, but I'm not sure if there's a specific (unambiguous) behavior we're trying to capture.  If there is, I'm happy to accommodate that (and edit the `equal?` cases while I'm at it).  If you're happy with the behavior but you want me to update the colorer and this function so that just the paren part goes through the colorer while the double quote and bar goes directly based on text, I can do that, too.

I should have time to work on whichever you think makes more sense in the next day or two.  What do you think?

rfindler:
I agree that figuring out matching when there aren't two separate characters requires significant extra work. Let's not hold up this pull request for that. 

For now, I'd be more than happy with going through the colorer to get the actual parenthesis and using that. Thank you!

bzyeung:
I think I have something ready for you to take a look at.  Do you want these updates squashed into one commit again, or would you prefer these be two separate commits?

gui-doc/scribblings/framework/color.scrbl
```diff
@@ -270,6 +270,14 @@
...
 
   }
+  @defmethod[(get-matching-paren-string [paren-str string?]
+                                        [get-side (or/c 'open 'close 'either) 'either])
+             (or/c string? #f)]{
+    Returns a string of a paren matching the other side of @racket[paren-str], if one exists
+    on the side indicated by @racket[get-side], or @racket[#f] if there is none.
+    If @racket[paren-str] includes any characters other than a single paren token (including
+    whitespace), this method returns @racket[#f].
+  }
```
rfindler: Probably good to connect this to the `pairs` argument of `start-colorer`?

gui-lib/framework/private/color.rkt
```diff
@@ -743,6 +745,21 @@ added get-regions
...
+    ;; String Symbol -> (U String False)
+    (define/public (get-matching-paren-string paren-str [get-side 'either])
+      (define sym (string->symbol paren-str))
+      (define checker
+        (case get-side
+          [(either) (λ (x) (member sym x))]
+          [(open) (λ (x) (equal? sym (cadr x)))]
+          [(close) (λ (x) (equal? sym (car x)))]
+          [else (raise-argument-error 'get-matching-paren-string
+                                      "(or 'open 'close 'either)"
```
rfindler: `or/c`, not `or`.

gui-test/framework/tests/racket.rkt
```diff
@@ -611,6 +621,45 @@
...
+      ))
+  (for ([s single-delete-scenarios]
+        #:unless (and (string=? "\"" open)
+                      (string=? "in-string" (car s))))
+    (let* ([before (cadr s)]
+           [after (string-append close (caddr s))]
+           [before-and-open (string-append before open)]
+           [result (list before after)])
+      (test-parens-behavior/full (format "~a-~a" (string-append open close) (car s))
+                                 before-and-open "" after #\backspace `(,result ,result)))))
```
rfindler: need some tests for the new method.

gui-doc/scribblings/framework/color.scrbl
bzyeung: Thanks, will do!

gui-lib/framework/private/color.rkt
bzyeung: Thanks!

gui-test/framework/tests/racket.rkt
bzyeung: Are the tests in color-text.rkt suitable if I move them over, but applied to an instance of racket:text% instead of color:text%?  Should I delete them from color-text.rkt, or is there somewhere else we have tests for (not racket-specific) color:text<%>?

rfindler:
Sorry, I got distracted when I was reading the code over: it looks like the current colorer doesn't actually get "\"" and "|" as things in the list of strings. Maybe that means we need a slightly different strategy somehow?

bzyeung:
I see several options, but none of them are perfect.  I think we agreed that doing all of parens, `"`, and `|` in `color:text<%>` is too much for now; I don't quickly see a way to do that without quite a bit of surgery.

One choice is whether we want to delete extra `"` or `|` by direct text manipulation or simply not do them at all.  Up to this point, my versions tried to do this, because we *add* `"` and `|` by direct text manipulation.  We could, of course, back this functionality out entirely, and decide that we're only going to delete "extra" *parens*, but that's unappealing because of the asymmetry between adding and deleting.  (Then again, I'm not sure it's possible to be "perfectly" symmetrical anyway.)

The other side is how we want to handle parens.  I agree that if we do it by direct text manipulation, to preserve some similar mechanisms with `"` and `|` removal, it's not entirely pretty, but I thought it was reasonable because there are so few cases for this feature; the intended behavior is only for empty pairs.  If we want to go through `color:text<%>`, one way to approach it without changing the interface might be to use `forward-match` and `classify-position` to find a parenthesis match one position apart.

The current version changes the interface by adding some type of `pair` field querying, which could be useful in a few other places in the racket text-mixin code.  We might even add more (in other PRs), to eventually eliminate `racket-paren:get-paren-pairs`, thus consolidating paren pairing between the colorer and racket module.  (We could also change `color:text<%>` in other ways, but I haven't looked deeply enough to see what might make sense.)

Maybe the answer is to push a text manipulation-only version or a `color:text<%>` version that does *not* change the interface for right now, and simultaneously open a new issue to hash out how we want to deal with parens more generally in the racket module?

bzyeung:
I came up with a couple more ideas, but maybe we should back up a bit to try to build some clarity here: can you be a bit more specific about what the issue with the existing strategies have been?  In particular, is there something that makes all of the existing proposals worse than, say, [bars and double quotes in `maybe-insert-brace-pair`](https://github.com/racket/gui/blob/2a76a975dc054bb2d9dab12e7d9cebd231d2357b/gui-lib/framework/private/racket.rkt#L1722)?

rfindler:
I mean that the new method you are adding makes sense for parentheses but I'm not seeing how it makes sense for the double quote and bar characters. Maybe I just missed something?

I'm thinking that the simple thing is best, namely adding the method but having it work only for the things that are already in paren pairs list and then, regrettably, just special casing double quote and bar, the way that they are already being handled elsewhere. 

It may make sense to have some more "centralized" knowledge of those two characters (somehow, whatever that might mean) in preparation for allowing the #lang line to customize them by specifying what the given language's matching delimiters are, but perhaps that's a better job in a separate issue.

(Does this make more sense?)

bzyeung:
Yes, it does!  Sorry; I thought this situation is what you were talking about with [this earlier comment](https://github.com/racket/gui/pull/162#issuecomment-603826452), so I was wondering whether your more recent remark meant I was missing a deeper flaw somewhere along the line.

I agree that doing the additional surgery to `color:text<%>` at some point is desirable, but I couldn't see a way to do it without significantly more work, and I'd want to give it more thought to make sure we did things intelligently.

The only other thought I had would be to start mucking with undo, but that would exhibit significantly different behavior.  (And isn't without its own potential mayhem.)

I'll open an issue to discuss how to consolidate handling of delimiters more generally.  In the meantime, are there any other edits you'd like for this current code?

bzyeung:
Sorry for the long fallow period, but I think the last to-do on this is to decide what to do with the tests; at present, I've got the tests located in two different files, which is probably not ideal.  Please let me know what I can do to get this part across the finish line.

(I hope to take a deeper dive into the broader questions raised in this conversation soon.)

rfindler:
I'm happy with the current state. I'll rebase to master.

PR #166, opened 2020-02-16
Use translations for common constants

alshopov:
Yes/No may lose shortcuts but these are currently non-translatable anyway.

Signed-off-by: Alexander Shopov <ash@kambanaria.org>


rfindler:
Two worries:

- the bigger one is that this introduces a dependency between racket/gui and string-constants. Probably have to get @mflatt to weigh in on that one

- the more minor one is that the `yes` and `no` string constants need the `&` in them (that's how the buttons get keyboard shortcuts on some platforms; the letter following the `&` gets an underline and is the keystroke to click the button).

rfindler:
Oh, sorry. I just looked at the diff, not your comment! How about just having a string constant `yes-in-a-button` where there is a note next to it saying what the `&` means (or similar)?

alshopov:
> ...How about just having a string constant `yes-in-a-button` ...
I will, though will suffix it with '-mnemonic'. Accelerators age generally available in many interface elements.

alshopov:
Land before https://github.com/racket/string-constants/pull/44

alshopov:
The checks fail becuase the updates to the string constants have not been merged.

rfindler:
The addition of the dependency from `racket/gui` to `string-constants` is problematic. Matthew conducted an experiment and reports that `string-constants` adds 8MB to a 66MB heap for `racket` with `racket/gui/base` loaded. For Racket CS it’s +14MB added to 153MB and these seem to be too much.

We could try to limit the amount of memory that a dependency would bring, or we could add optional, keyword arguments to these functions and then have framework/gui-utils supply translations of those arguments (exporting functions to be called in place of `get-file` et al).

alshopov:
> The addition of the dependency from racket/gui to string-constants is problematic.

Did I really do that? `gui-lib/info.rkt ` already contained `["string-constants-lib" #:version "1.24"]`, I just bumped it up to `1.33`. Are there so many new messages and are they loaded at the same time to take up this additional memory? I though that there is only one translation loaded at a single time so even though some more need more memory - that will be offset by the removal of the English versions.
While I did add `string-constants` to required of `gui-lib/mred/private/messagebox.rkt`, `gui-lib/mrlib/terminal.rkt` already did that and there are many more other files requiring the library so this package has already been used - `git grep '[(]string-constant' | wc -l` in the repo gives more that 360 usages. What am I missing in the picture and how can I improve things?

rfindler:
The pkg level dependency was there; the module level one wasn't, I believe.

Robby

alshopov:
> the module level one wasn't, I believe.

I have not added anything on module level. The whole change was bumping the dependency at pkg level and using 3 constants. Perhaps there is something trivial I am overlooking.

rfindler:
The issue is the addition of the new line 7 in gui-lib/mred/private/messagebox.rkt. That makes the memory use of the programs that Matthew reported jump and what I mean by "adding a dependency" from racket/gui to string-constants.

Does this make more sense?

rfindler:
I don't think we can merge this one in its current state.

We could change things by some how parameterizing `message-box` and friends to internationalize them and then have properly internationalized versions available via the framework.

rfindler:
Would it be horrible to add optional arguments to `message-box` (and friends) but where their default values were determined by some mutable state and then have the framework gui library mutate that state (using string constants)?

We could document this in the docs for `message-box` (since a dependency in the documentation is okay), telling people that something fishy is going on.

alshopov:
I have no firm opinion on that. I guess it would be ok. Smooth running of
Racket is the priority.

На нд, 27.09.2020 г. в 21:46 Robby Findler <notifications@github.com>
написа:

PR #167, opened 2020-02-16
fixes auto-parens when typing quote or bar near comments/strings

### LINKED ISSUE #164: With auto-parens, " or | near strings or comments is not correctly handled
With automatic parentheses on, when `"` or `|` are typed at the end of a line comment, a second character gets added.  However, as both characters are now part of the comment, the character isn't smart-skipped by typing it again.  In a related problem, when typing those characters at the beginning, "outside" the comment, they do *not* add the extra character.  These are both in contrast to the behavior with parens, so I think this qualifies as a bug.

Given that I've just been playing with similar code in the same area while working on new feature #162, I'm happy to take a swing at this, unless this behavior is actually intended.

bzyeung:
Fixes #164.

### DIFF
--- a/gui-lib/framework/private/racket.rkt
+++ b/gui-lib/framework/private/racket.rkt
@@ -1692,14 +1692,14 @@
           (match cur-token
             [(or 'error #f) (insert-brace-pair text open-brace close-brace 'error)]
             ['constant (insert-brace-pair text open-brace close-brace 
-                                       (λ(t) (not (eq? t 'constant))))]
+                                          (λ (t) (not (equal? t 'constant))))]
             [(or 'symbol 'comment)
              (cond
                [(and c (char=? #\| open-brace) (string=? c "|"))   ;; smart skip
-                   (send text set-position (+ 1 (send text get-end-position)))
-                   (define d (immediately-following-cursor text))
-                   (when (and d (string=? d "#"))   ; a block comment?
-                     (send text set-position (+ 1 (send text get-end-position))))]
+                (send text set-position (+ 1 (send text get-end-position)))
+                (define d (immediately-following-cursor text))
+                (when (and d (string=? d "#"))   ; a block comment?
+                  (send text set-position (+ 1 (send text get-end-position))))]
                [(in-position? text '(comment)) (send text insert open-brace)]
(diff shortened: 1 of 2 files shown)


bzyeung:
Regarding style/etiquette, I noticed that some of the indentation was nonstandard and updated that as part of the commit.  However, I tried to be minimal with respect to any other edits.  I'm happy to incorporate any feedback along the lines of whether I should have separated the indentation from the other edits or whether I should have tried to do any additional editing while I was in there.  Thanks!

rfindler:
Just looking at the code, I think this is good (I didn't confirm that the test cases actually cover all the new behavior.)

Very minor, but since you brought it up, I think it would be great if you didn't change the indentation unless you had to actually change the corresponding line but if that's too annoying, I understand. Relatedly, there should be a space after the lambda and before the open paren. Also, use `equal?`, not `eq?` here. These aren't necessary to accept the patch, tho!

bzyeung:
Thanks for the quick turnaround.  I'm happy to make those edits and resubmit the PR if that's preferred.  It's not too annoying for me to leave the old indentation.  Should something like that have been a separate commit/PR, or is it customary to just leave those issues alone entirely?

As for `equal?` versus `eq?`, should I also change the existing use of `eq?` (in the `'constant` case) to `equal?` while I'm at it?

rfindler:
Leaving them alone or putting them into a separate, housekeeping commit is best in my opinion. For `equal?` vs `eq?` that's another housekeeping issue so I would just not put any new ones in the commit that fixes the bug. Feel free to just force push to the PR.

bzyeung:
I've backed out the indentation changes to lines that were not otherwise edited.  Should a housekeeping commit go in as a second commit in the same PR, or is that better as a second PR after this one is merged?

rfindler:
If the two commits had been squashed into one, I would have merged the PR :). But since they aren't yet, feel free to do it either way.

bzyeung:
Heh.  Sorry!  I thought that if I squashed them together, that would defeat the purpose of separating the housekeeping changes to begin with.  (I've never been on the other side of those.)  So you can still see the separate changes if I were to have done that?

rfindler:
Oh, maybe I confused myself? Anyway, I've merged it now.

Thanks!

PR #172, opened 2020-02-25
Fixes auto-parens for double quote before later double quote string

### LINKED ISSUE #171: Automatic double-quote insertion broken
In the current git master, it appears that insertion of double-quotes is broken when inserting
double-quotes before other double-quotes.

How to reproduce:
1) Start DrRacket, select the "Automatic Parentheses & Quotes" option in editing>general.
2) Open a new buffer
3) hit return three times, type a double-quote. See two double-quotes inserted, as expected.
4) hit the up arrow twice. Type a double quote. See only *one* double-quote inserted, and 
syntax highlighting become angry red.

I believe this error does not affect the v7.6 release.

bzyeung:
Fixes #171.


rfindler:
Thanks!

PR #173, opened 2020-03-14
fix paren insertion at eol of line comment

Metaxal:
This bug fix follows a discussion with Robby on slack #drracket. The bug was that when typing a closing parenthesis at the end of line of a line comment, smart-skip may skip to the next closing parenthesis on the next line, because the lexer believes it is already outside the line comment
(this behaviour is correct for nested comments `#|...|#` but not for line comments `; ...`). See the two new test cases.

The current proposal is a tradeoff between simplicity, bug fix, and specificity to the racket language.

Note that for homogeneity (and robustness), I also changed in the same function the previous check for beginning of strings and beginning of comments (where characters are actually not inserted in the string or comment) to follow the same pattern I use for the bug fix, which is to use the token start position rather than classifying left/right, which I believe is a more proper and robust way to check these cases. 

All tests from raco test in the gui root dir pass.

Happy to discuss this further, and thank you for your help on this.

### DIFF
--- a/gui-lib/framework/private/racket.rkt
+++ b/gui-lib/framework/private/racket.rkt
@@ -1434,15 +1434,31 @@
 ;; classify-position characterizes it so).
 (define (in-position? text sym-list)
   (define selection-start (send text get-start-position))
-  (define first-type (send text classify-position selection-start))
-  (define final-type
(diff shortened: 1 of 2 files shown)


rfindler:
Thanks!

PR #181, opened 2020-05-03
Clean up misspelled string/text handling in `add-colorings`

gfbee:
A prelude to modifying `add-colorings` to support https://github.com/racket/drracket/issues/367 .

Regardless of whether we touch that implementation, we could consider making `add-colorings` public, and then specialized sub-coloring could be mixed in. I happen to be working on some sub-colorings myself, which would benefit from having that api.

### DIFF
--- a/gui-lib/framework/private/color.rkt
+++ b/gui-lib/framework/private/color.rkt
@@ -441,26 +441,27 @@ added get-regions
                                  sp
                                  query-aspell))
             (for ([spell-info (in-list spell-infos)])
-              (case (car spell-info)
+              (define maybe-suggestions (car spell-info))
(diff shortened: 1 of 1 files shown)


rfindler:
As we discussed briefly on slack, I'm not convinced this makes sense as a prelude. Also we would need documentation and tests if we were to add a new api.

gfbee:
Just want to emphasize that this cleanup is orthogonal to whether we extend the api.
A cleanup is a low-priority change, with a bit of a risk/benefit concern, but if someone is about to mimic that code I'd suggest they look at this version to mimic (at which point I can argue that organizing both similarly has a better risk/benefit tradeoff).

PR #187, opened 2020-05-10
Fix spell checking

gfbee:
New approach that Identifies spelling suggestions for clients by rechecking the token at the requested position, rather than trying to track shifting regions created when misspellings are colored.

The former implementation was buggy when misspelled regions changed token type or were edited without completely being replaced. A correct implementation following its approach would still be tricky, and prone to efficiency costs that the new approach shifts completely to when the user requests the location of a misspelled word or its suggestions.

The user now only pays for spell check features (find misspelled word, request suggestions) when used (and which is rare compared to general editing and recoloring). Moreover, the overhead is negligible relative to the user's interaction time with the feature (context switch, menu selection or keypress, consideration of and reaction to the result).

### DIFF
--- a/gui-lib/framework/private/color.rkt
+++ b/gui-lib/framework/private/color.rkt
@@ -447,19 +447,23 @@ added get-regions
       (or (and (equal? type 'string) spell-check-strings?)
           (and (equal? type 'text) spell-check-text?)))
     
+    ; This implementation (along with changes in the rest of the file) fixes spell-checking,
+    ;  which was only working before when used in a stereotyped way.
+    ;
+    ; Some notes about the properties of this implementation ...
+    ;
     ; The 7.7 docs specify that this method produces #f if checking is off (as well as if there are
     ;  no suggestions), although that behavior is unlikely to be usefully relied on in that case.
-    ;  The new implementation preserves that behavior.
+    ;  That behavior is preserved.
     ;
-    ; The new implementation relies on the colorer not being stopped, and produces #f if it is.
+    ; The method needs information from the colorer, and produces #f if the colorer is stopped.
     ;
-    ; The previous implementation attempted to initiate and then maintain a set of misspelled regions
-    ;  based on an initial coloring, but was fundamentally broken, unless used in a stereotyped way.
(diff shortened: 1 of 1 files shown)


rfindler:
Does this lose the coloring for misspelled words?

gfbee:
It shouldn't, but if you're finding it does I'll get more serious about my aspell mocking while testing.

The coloring for misspelled words happens at https://github.com/racket/gui/blob/12a2672b4cbf3bd45e295c38c6c74b21f60dcc96/gui-lib/framework/private/color.rkt#L444 which collapses a former `case` to an `if` since the branches are the same now except for choice of a single value.

Without the region tracking, the entire body of the old `add-coloring/spell` is just a call to `add-coloring`. Since all token coloring passed through there due to a historical attempt to remove non-misspelled regions, it made sense to do the coloring while there. Now that's a pointless indirection, which I've inlined everywhere: https://github.com/racket/gui/blob/12a2672b4cbf3bd45e295c38c6c74b21f60dcc96/gui-lib/framework/private/color.rkt#L444-L449

rfindler:
Thank you for this. I've read the code and understand what you mean now. I agree this is a better strategy!

I don't understand this comment, however: "It also had the property (arguably a bug, although mild and obscure) of requiring the style list have a style for misspelled colors. That incidental property isn't preserved here." It looks like the misspelled color style is still needed?

rfindler:
I'm happy to merge there @gfbee but I would prefer to remove the comment that doesn't seem to be true first. Or, if it is true, I'd like to understand why! :)

gfbee:
Let me know if the new comment has anything unclear or apparently incorrect.

There's a lot of implicit legacy knowledge in a code base as large and old as this, and the previous comment was partly to highlight for you anything that you might realize has more significance than I could know. It's “just” spell checking, and the changes have little effect on anything else, but the more explicit we can make the design decisions and their effects, the more easily I can contribute.

rfindler:
This is great, thanks! I see what my confusion was before. I'll squash and rebase.

rfindler:
(I also took the liberty of removing the now-unused require before pushing.)

PR #190, opened 2020-05-22
Update gui-dynamic-require scribblings to reflect the necessity of us…

dzoep:
…ing dynamic-require beforehand.

As discussed at #Racket, dynamic-require has to be used before gui-dynamic-require can do anything.


mflatt:
Thanks for the patch! I took the liberty of revising it (I hope that's ok) to better explain the original intent, make the connection to `gui-available?`, and correct some details details before merging.

Issue #192, opened 2020-06-14
Call to set-padding on text% hangs


aymanosman:
Minimal example

```
#lang racket/gui

(define frame
  (new frame% [label "my frame"] [width 400] [height 300]))

(define canvas (new editor-canvas% [parent frame]))
(define text (new text%))

(send canvas set-editor text) ;; won't hang if this is called below

(send text auto-wrap #t)
(send text set-padding 10 10 10 10) ;; hangs here

;; (send canvas set-editor text) ;; won't hang if this is called here

(send frame show #t)
```

Tried on Racket 7.6 and 7.7 [3m].

alex-hhh:
I can confirm that this is happening for me too.  I think the problem is that the `editor-canvas%` has zero dimensions until the frame is shown, and the call to `set-padding` is somehow affected by that.  Adding a call to `reflow-container` after the canvas is created also "fixes" the problem:

```
(send frame reflow-container)
```

PR #195, opened 2020-06-20
Ensure that the canvas width is sufficient for reflowing lines (fix for #192)

alex-hhh:
This is a fix for #192, and the example in that issue works fine with this change.

## Explanation

Calling `set-padding`, will result in a call to `recalc-lines` (via calls to `needs-update` -> `continue-update` -> `redraw`).

Before the `text%` objects is shown and the canvas dimensions are computed, the `max-width` will be 3, and when the left, right padding and `CURSOR-WIDTH` are subtracted from it, the remaining width will be negative.  As a result of this, `mline-update-flow` will always return `#t`, so the loop never terminates.

I changed the code to calculate the final width and checking that result to be positive.

## Things I still don't understand

* I am unsure why `max-width` was 3 before the canvas was shown (using the example from #192), I was expecting it to be 0
* ~I could not find the definition of `mline-update-flow` using a GitHub search, so I don't know where that function is defined.~  -- I found the definition.  It is in mline.rkt under the name `update-flow`.
* The check for the width is still to be greater than 0, which is what the original code did, but I am unsure if the loop will terminate when the width is a very small value, for example when someone sets a big padding, such that the remaining width is 1.   I think an improvement would be not only to check if `mline-update-flow` needs further re-flows, but also if anything changed since the last re-flow (that is, to see if a re-flow is in fact possible at all).

Still, at least the case described in #192 is fixed by this pull request.

### DIFF
--- a/gui-lib/mred/private/wxme/text.rkt
+++ b/gui-lib/mred/private/wxme/text.rkt
@@ -4935,14 +4935,14 @@
               (loop (mline-next line)))))
         
         (let ([-changed?
-               (if (max-width . > . 0)
-                   (let ([wl? write-locked?]
-                         [fl? flow-locked?])
-                     ;; if any flow is updated, snip sizing methods will be called
-                     (set! write-locked? #t)
-                     (set! flow-locked? #t)
-                     
-                     (let ([w (- max-width padding-l padding-t CURSOR-WIDTH)])
+               (let ([w (- max-width padding-l padding-t CURSOR-WIDTH)])
+                 (if (w . > . 0)
+                     (let ([wl? write-locked?]
+                           [fl? flow-locked?])
+                       ;; if any flow is updated, snip sizing methods will be called
+                       (set! write-locked? #t)
+                       (set! flow-locked? #t)
+                       
                        (let loop ([-changed? #f])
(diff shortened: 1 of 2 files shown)


sorawee:
FWIW, here's `mline-update-flow`: https://github.com/racket/gui/blob/master/gui-lib/mred/private/wxme/mline.rkt#L933

alex-hhh:
Hi @samth or @mflatt, now that 7.8 has been released, can this pull request be merged?  Thanks, Alex.

samth:
I think @mflatt needs to be the one to review this -- I don't know this code at all.

mflatt:
Sorry that I I overlooked this one!

I'm happy to merge, but can you turn the issue into a test case, probably just as a module in "gui-test/tests/gracket"? A variant of the program that ends fairly quickly as opposed to looping forever would be good enough as a test.

alex-hhh:
Hi @mflatt, I added a test case based on the reported issue -- the test will open a window and close it, when the bug is fixed, but will just hang if the bug is present.  Is this what you had in mind?

mflatt:
Yes - thanks!

PR #205, opened 2020-11-27
Migrate from Travis to GHA

sorawee:
- Travis is no longer a viable option.
- Even if it is, we are using travis-ci.ORG, which will be out of service next month.

This PR switches from Travis to GHA.

Note that the current Travis config has been failing for 3 months, with the error `ls: cannot access '/home/travis/.racket/download-cache': No such file or directory`, so I simply remove the problematic step.


samth:
We should make a github action that automates this, but this is great!

PR #213, opened 2021-01-27
Updated documentation of default snip insert location

bmitc:
The default insert location of a snip on a pasteboard was incorrectly documented as (0,0), whereas the behavior is that the snip is inserted at the center of the pasteboard. See https://github.com/racket/gui/blob/7c3533424f01fc667d5ec0f7598614f367892eac/gui-lib/mred/private/wxme/pasteboard.rkt#L712

### DIFF
--- a/gui-doc/scribblings/gui/pasteboard-class.scrbl
+++ b/gui-doc/scribblings/gui/pasteboard-class.scrbl
@@ -682,7 +682,7 @@ Inserts @racket[snip] at @techlink{location} @math{(@racket[x],
  @racket[before]. (@|seesniporderdiscuss|) If @racket[before] is not
  provided or is @racket[#f], then @racket[snip] is inserted behind all
  other snips. If @racket[x] and @racket[y] are not provided, the snip
- is added at @math{(0, 0)}.
+ is added at the center of the pasteboard.
 
 }
 


bmitc:
Actually, I'm not for sure what to do with this one now. I have found that the behavior is inconsistent. For example, see the following code.

```
#lang racket/gui

(define node-snip-class
  (make-object
      (class snip-class%
        (super-new)
        (send this set-classname "node-snip"))))

(send (get-the-snip-class-list) add node-snip-class)

(define (set-extent-others others)
  (for/list ([other others])
    (when other (set-box! other 0.0))))

(define node%
  (class snip%
    (init-field size color)
    (super-new)
    (send this set-snipclass node-snip-class)

    (define/override (get-extent dc x y width height . others)
      (when width (set-box! width size))
      (when height (set-box! height size))
      (set-extent-others others))

    (define/override (draw dc x y . other)
      (send dc set-pen "Black" 3 'solid)
      (send dc set-brush color 'solid)
      (send dc draw-rounded-rectangle x y size size 7))
    ))

(define diagram (new pasteboard%))

(define frame (new frame%
                   [label "Diagram"]
                   [width 500]
                   [height 500]))

(define canvas (new editor-canvas%
                    [parent frame]
                    [style '(no-hscroll no-vscroll)]
                    [horizontal-inset 0]
                    [vertical-inset 0]
                    [editor diagram]))

(send diagram insert (new node% [size 50] [color "Pale Green"]))

(send frame show #t)

(send diagram insert (new node% [size 50] [color "Pale Turquoise"]))
```

Note how whether the node is inserted before or after the frame is shown determines where the node is automatically inserted. Also, note how the nodes aren't entered exactly at (0,0) or the center.

image


Any thoughts on what to do here?

alex-hhh:
The snip is always inserted in the center of the pasteboard, but the center is not fixed, as it depends on the size of the canvas:

* before the frame is shown, the full layout of the components is not calculated yet, so the pasteboard has a default size (I believe this would be a size of 0, although the insets are might already be factored in).  So, your first snip is inserted at the center of the pasteboard when its size was 0.  
* After the frame is shown, the components sizes are calculated, so the editor canvas has the size you expect and the second snip is inserted, again in the center of the pasteboard, but this time the pasteboard as a different size and therefore a different center.

You cannot rely of the size of any components before the container (the frame, in your case), has calculated its layout.  This happens automatically when the frame is shown or resized, but you can also trigger a re-calculation using `reflow-container`.  For example, if you add `(send frame reflow-container)` before the first snip is inserted, it too will be in the expected center.

I would not recommend relying on a specific insert location, if you don't specify one.  The "center" will also change if the user resizes the window, but the already inserted snips will remain where they are.

----

This widget behavior is normal for most of the GUI systems I am familiar with.  It is not specific to Racket GUI.

bmitc:
Thanks for that additional detail that explains the situation for when the nodes are drawn before showing the frame. I personally think it's weird that this occurs, because `show` is doing more than show, but your explanation makes sense. The [documentation for show](https://docs.racket-lang.org/gui/window___.html#%28meth._%28%28%28lib._mred%2Fmain..rkt%29._window~3c~25~3e%29._show%29%29) does not mention this.

I think the documentation should be further updated then to help provide a summary of this detail instead of just my correction saying the snips are inserted at the center instead of (0,0).

rfindler:
Sorry about the delay here. The additions to the docs sound helpful, indeed.

PR #214, opened 2021-02-12
guard mouse-event

bdeket:
guard from using mouse-event methods on a key-event

close [drracket/issues/466](https://github.com/racket/drracket/issues/466)


rfindler:
Thank you! I had tested that same change on my windows machine but didn't quite get the push done before seeing this.

PR #215, opened 2021-02-20
Handle image-markup

mikesperber:
I just added `image-markup` to `simple-tree-text-markup/data`. This is the rendering support for that.

(I'm not really happy with `record-dc-datum->bitmap` as there's no predicate for the `record-dc%` data.)

### DIFF
--- a/gui-lib/framework/private/text-port.rkt
+++ b/gui-lib/framework/private/text-port.rkt
@@ -3,9 +3,11 @@
          "text-sig.rkt"
          (prefix-in base: racket/base)
          simple-tree-text-markup/data
+         (only-in simple-tree-text-markup/construct markup-transform-image-data)
          racket/unit
          racket/class
          racket/match
+         racket/draw
          mred/mred-sig
          mrlib/interactive-value-port
          (prefix-in image-core: mrlib/image-core))
@@ -17,7 +19,8 @@
           (except text-mixed-in-classes^ keymap%)
           [prefix icon: framework:icon^]
           [prefix editor: framework:editor^]
-          [prefix srcloc-snip: framework:srcloc-snip^])
+          [prefix srcloc-snip: framework:srcloc-snip^]
+          [prefix text: text-misc^])
   (export text-port^)
 
   (define wide-snip<%>
@@ -68,11 +71,12 @@
 
   ;; class for snips embedded in markup
   (define markup-text%
-    (wide-snip-mixin
-     (basic-mixin
-      (editor:standard-style-list-mixin
-       (editor:basic-mixin
-        text%)))))
+    (text:foreground-color-mixin
+     (wide-snip-mixin
+      (basic-mixin
+       (editor:standard-style-list-mixin
+        (editor:basic-mixin
+         text%))))))
   
   (define-struct peeker (bytes skip-count pe resp-chan nack polling?) #:inspector (make-inspector))
   (define-struct committer (kr commit-peeker-evt done-evt resp-chan resp-nack))
@@ -523,7 +527,9 @@
              (do-insertion txts #f)
              (if async? (thread (λ () (sync signal))) (sync signal)))
            #f)))
-    
+
+      (inherit line-start-position position-line)
+      
       ;; do-insertion : (listof (cons (union string snip) style-delta)) boolean -> void
       ;; thread: eventspace main thread
       (define/private (do-insertion txts showing-input?)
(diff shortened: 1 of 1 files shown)


rfindler:
I'm not sure that the `((is-a? data snip%) data)` is a good idea.

I'm not sure what guarantees the record-dc gives when given bogus data but it stands to reason that it should offer either a predicate or a promise that an exception is raised at a specific moment. In other words, if neither of those is true, we should consider it a bug in record-dc.

mikesperber:
Revised the patch as per e-mail discussion.

Two issues remain:

- If I explicitly set the style on the inner snips, that destroys the
  styling already done within those snips - notably the underlining for
  srclocs. (See `FIXME`.)

- The frame around `framed-markup` doesn't change color in dark mode.

gui-lib/framework/private/text-port.rkt
```diff
@@ -538,50 +544,159 @@
...
+                   ((snip-special? data)
+                    (define snip (snip-special->snip data))
+                    (if (exn:fail? snip)
+                        (markup->snip (image-markup-alt-markup markup) style #f)
+                        snip))
+                   (else
+                    (markup->snip (image-markup-alt-markup markup) style #f)))))
+             
+             (define (record-dc-datum->bitmap datum width height)
+               (with-handlers ((exn? (lambda (e) #f)))
```
rfindler, line 595: This should be `exn:fail?` at a minimum so we don't handle `exn:break?` (and perhaps it should be tightened to wrap around the calls to the recorded-datum functions).

gui-lib/framework/private/text-port.rkt
```diff
@@ -538,50 +544,159 @@
...
+                   ((is-a? data snip%) data)
+                   ((snip-special? data)
+                    (define snip (snip-special->snip data))
+                    (if (exn:fail? snip)
+                        (markup->snip (image-markup-alt-markup markup) style #f)
+                        snip))
+                   (else
+                    (markup->snip (image-markup-alt-markup markup) style #f)))))
+             
+             (define (record-dc-datum->bitmap datum width height)
```
rfindler, line 594: Why not just draw into the dc instead of creating a bitmap?

gui-lib/framework/private/text-port.rkt
```diff
@@ -538,50 +544,159 @@
...
+                           (srcloc-snip:select-srcloc (srcloc-markup-srcloc srcloc-markup))))
+                   (send text change-style
+                         (make-object style-delta% 'change-underline #t)
+                         start end #f))))
+             
+             ; like for-each, but with a thunk that gets called between elements
+             (define (for-each/between proc between list)
+               (let loop ((list list))
+                 (cond
+                   ((null? list) (values))
```
rfindler, line 618: Maybe `(void)` instead of `(values)`?

gui-lib/framework/private/text-port.rkt
```diff
@@ -1004,6 +1058,37 @@
...
+              (regexp-match? #rx"pict3d%" str))
+          (send value copy)]
+         [else
+          (define special (make-snip-special (send value copy)))
+          (unless (snip-special-name special)
+            (error 'encode-image-data "unencodable image snip" value))
+          special])]
+      [(is-a? value bitmap%) (make-object image-snip% value)]
+      [else
+       (with-handlers ((exn? (lambda (e) #f)))
```
rfindler, line 1085: `exn:fail?` here too.

gui-lib/framework/private/text-port.rkt
```diff
@@ -1004,6 +1058,37 @@
...
+         (let ((proc (recorded-datum->procedure value))
+               (bitmap (make-object bitmap% width height)))
+           (let ((dc (new bitmap-dc% [bitmap bitmap] )))
+             (proc dc)
+             (make-object image-snip% bitmap))))]))
```
rfindler, line 1090: Also I think we don't want to create a bitmap here.

gui-lib/framework/private/text-port.rkt
```diff
@@ -538,50 +544,159 @@
...
+                      [snip (new editor-snip% [editor text] [with-border? framed?])])
+                 (send snip use-style-background #t)
+                 (send text set-styles-sticky #f)
+                 (insert-markup markup text style #f)
+                 (send text lock #t)
+                 snip))
+
+             (define (insert-markup markup text style inline?)
+               ; FIXME: Mike doesn't know how to apply style here without destroying
+               ; the style set by, say, srcloc-markup
```
rfindler, line 559: Does it work to use a `style-delta%`?

gui-lib/framework/private/text-port.rkt
mikesperber: Which dc? Don't I need a bitmap to create a `bitmap-dc%`?

gui-lib/framework/private/text-port.rkt
mikesperber: Done.

gui-lib/framework/private/text-port.rkt
mikesperber: Done.

gui-lib/framework/private/text-port.rkt
mikesperber: The `style` is passed in from the outside - I didn't create it.

gui-lib/framework/private/text-port.rkt
mikesperber: Done. (Even though I really like `values` better here, so that no one gets tempted to do anything with the return value.)

gui-lib/framework/private/text-port.rkt
mikesperber: I think these are `(get-out-style-delta)` etc., suggesting that they already are style deltas, but I have a hard time understanding how the names get turned into style objects.

gui-lib/framework/private/text-port.rkt
rfindler: I'm saying that you could save the recorded-datum and then later, just use it to draw.

I suppose this means that we'd need a record-dc-snip% (which isn't entirely trivial, I am now realizing but perhaps not that bad).

gui-lib/framework/private/text-port.rkt
rfindler: Thanks. `(void)` seems to be the convention elsewhere is all so it might confuse. I wonder which would work better for useful error messages when a bug happens.

gui-lib/framework/private/text-port.rkt
mikesperber: Fixed this as per our e-mail discussion by calling `set-style` on the snip.

PR #222, opened 2021-03-27
Add option in  that controls Home and End. Default unchanged.

soegaard:
This PR adds a new option in "Editing | General Editing".

The conventions on macOS differ from Windows/Linux when it comes 
the handling of the Home and End keys. On macOs they are move
to the beginning and end of the document, on the other systems to
the beginning and end of the line.

If the new option is checked, the macOS convention is used.
The option also affects shift+Home/End.

The default value is set of false. Should this default to true on macOS?

image

### DIFF
--- a/gui-lib/framework/private/keymap-global.rkt
+++ b/gui-lib/framework/private/keymap-global.rkt
@@ -520,7 +520,38 @@
                         "delete-next-character"
                         "delete-previous-character")
                     edit event #t)))]
-         
+         [home-key          
+          (λ (edit event)
+            (let ([kmap (send edit get-keymap)])
+              (send kmap call-function
+                    (if (preferences:get 'framework:home/end-to-beginning/end-of-document?)
+                        "beginning-of-file"
+                        "beginning-of-line")
+                    edit event #t)))]
+         [end-key
+          (λ (edit event)
+            (let ([kmap (send edit get-keymap)])
+              (send kmap call-function
+                    (if (preferences:get 'framework:home/end-to-beginning/end-of-document?)
+                        "end-of-file"
+                        "end-of-line")
+                    edit event #t)))]
+         [shift-home-key          
+          (λ (edit event)
+            (let ([kmap (send edit get-keymap)])
+              (send kmap call-function
+                    (if (preferences:get 'framework:home/end-to-beginning/end-of-document?)
+                        "select-to-beginning-of-file"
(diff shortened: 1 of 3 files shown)


soegaard:
Note that this PR relies on 

    https://github.com/racket/string-constants/pull/51
    
which add the string constant missing in the test above.

rfindler:
I think it would be better if it was clear what both states of the checkbox mean. One option would be to have a more verbose checkbox; another would be to use radio buttons with two strings. 

Probably it is good to add some test cases for this.

The documentation in DrRacket also needs a tweak for this change.

There should be a bump to the version dependency with this too.

soegaard:
> I think it would be better if it was clear what both states of the checkbox mean. 
> One option would be to have a more verbose checkbox; another would be to use radio buttons with two strings.

Ideally I'd like to add a tooltip to this check box.

Alex Harsanyi has a nice tooltip-mixin, but it would be better to have similar functionality directly in the framework.
https://docs.racket-lang.org/gui-widget-mixins/index.html?q=gui-widget-mixins

In the mean time:
    Now:              "Home and End go to the beginning or end of the document"
    Suggestion:  "If checked, Home and End keys go to beginning and end of the document. 
                            If unchecked, they go to the beginning and end of the line."

I think, visually it is better to keep a check box - there are no other radiobuttons in that dialog.
Also, there are only two options. [1]

> Probably it is good to add some test cases for this.
Is there an existing test that can be mimicked?

> The documentation in DrRacket also needs a tweak for this change.

I looked at https://docs.racket-lang.org/drracket/Keyboard_Shortcuts.html
Now, it doesn't mention Home and End. Is this the page, you had in mind?

> There should be a bump to the version dependency with this too.
Ok.

[1] "Consider a checkbox when the user needs to choose between two states."
https://developer.apple.com/design/human-interface-guidelines/macos/buttons/radio-buttons/

rfindler:
I'll have to ask the release managers about gui-widget-mixins as it would be a new pkg in the distribution but I like the idea of a tooltip.

The page I had in mind is [this one](https://docs.racket-lang.org/drracket/prefs-explanation.html), sorry about that.

rfindler:
Hi @soegaard : let's go ahead and add that dependency to have a tooltip for this checkbox item. But let's hold off until next week to merge this pull request so that it'll happen only in the release after the upcoming one (just to avoid trouble).

PR #226, opened 2021-04-23
Render number-markup.

mikesperber:


gui-lib/framework/private/text-port.rkt
```diff
@@ -590,7 +593,25 @@
...
+                  (string-append (if prefix? "#e" "") (number->string number))]
+                 [else
+                  (case fraction-view
+                    [(mixed improper)
+                     (define snip (number-snip:make-fraction-snip number prefix?))
+                     (send snip set-fraction-view fraction-view)
+                     snip]
+                    [else ; decimal
+                     (number-snip:make-repeating-decimal-snip number prefix?)])]))
+
```
rfindler, line 614: Is this code a copy of some code elsewhere?

mikesperber: > Is this code a copy of some code elsewhere?

No. I took inspiration from here:

https://github.com/racket/drracket/blob/917c6abdd3b146cfd85b08fdc7ad39f48cd834e3/drracket/drracket/private/language.rkt#L433

(Sorry, did not see this earlier.)

mikesperber:
@rfindler Revised the pull request as per e-mail discussion to factor out number-snip generation into the framework. I also have a patch for DrRacket ready to go, once the dust settles on this one.

In the docs, I expected a link to `pretty-print-size-hook` to be generated, but it doesn't - don't know why. (There is a `for-doc` require for `racket/pretty`.)

rfindler:
> In the docs, I expected a link to `pretty-print-size-hook` to be generated, but it doesn't - don't know why. (There is a `for-doc` require for `racket/pretty`.)

Maybe it needs to be a `for-label` inside the `for-docs`?

mikesperber:
Revision as per our discussion.

mikesperber:
> > In the docs, I expected a link to `pretty-print-size-hook` to be generated, but it doesn't - don't know why. (There is a `for-doc` require for `racket/pretty`.)
> 
> Maybe it needs to be a `for-label` inside the `for-docs`?

Yup, did that.

PR #228, opened 2021-06-12
list-box: fix cell higlighting on macOS

Bogdanp:
Here's what the following program looks like

```racket
#lang racket/gui

(define f (new frame% [label "LB"]))
(define l (new list-box% [parent f] [label #f] [choices '("a" "b")]))
(send f show #t)
```

before this change:

Screenshot 2021-06-12 at 10 43 25

and after:

Screenshot 2021-06-12 at 10 42 43


mflatt:
Thanks!

PR #229, opened 2021-06-15
Increase view used while printing

plane:
Fix for https://github.com/racket/drracket/issues/147

When printing the definitions area in DrRacket, it makes sense to have
the viewable range be essentially infinite, so all of the line numbers
in the file are always printed.  Currently it is hard-coded at a size
of 10000x10000, which turns out to not be high enough.

This patch changes the size to a very large number (1e50), and in my
tests it fixes the linked bug.


mflatt:
Thanks!

PR #231, opened 2021-07-03
fix clipping rect for nested text

### LINKED ISSUE #225: editor-snip% doesn't interact with horizontal-inset properly
```racket
#lang racket/gui

(define f (new frame% [label "test"] [width 320] [height 240]))
(define t (new text%))
(define c (new editor-canvas% [parent f] [editor t] [horizontal-inset 100]))

(send t insert "111111111111111111111\n")

(define t2 (new text%))
(define sn (new editor-snip% [editor t2]))
(send t2 insert "111111111111111111111\n")
(send t insert sn)

(send f show #t)
```
results to
图片

yjqww6:
fix #225.

### DIFF
--- a/gui-lib/mred/private/wxme/text.rkt
+++ b/gui-lib/mred/private/wxme/text.rkt
@@ -23,7 +23,8 @@
                   editor-snip%)
          "wordbreak.rkt"
          "stream.rkt"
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #232, opened 2021-07-09
Bump gui-lib dependency to 8.2.0.1

shhyou:
The improvement of the layering support (16147c5e5a) depends
on the new API `get-gui-bin-extra-search-dirs` introduced
in Racket v8.2 (racket/racket@dfbb7040aa).


mflatt:
Thanks!

PR #238, opened 2021-08-19
message%: add support for text colors

Bogdanp:
The CI failure is expected since the GTK change depends on a change to `draw-lib`.

### DIFF
--- a/gui-doc/scribblings/gui/message-class.scrbl
+++ b/gui-doc/scribblings/gui/message-class.scrbl
@@ -16,7 +16,7 @@ A message control is a static line of text or a static bitmap. The
                                (is-a?/c panel%) (is-a?/c pane%))]
                  [style (listof (or/c 'deleted)) null]
                  [font (is-a?/c font%) normal-control-font]
-                 [color (or/c #f (is-a?/c color%)) #f]
+                 [color (or/c #f string? (is-a?/c color%)) #f]
                  [enabled any/c #t]
                  [vert-margin spacing-integer? 2]
                  [horiz-margin spacing-integer? 2]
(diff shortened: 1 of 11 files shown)


gui-doc/scribblings/gui/message-class.scrbl
```diff
@@ -65,5 +74,20 @@ The same as @xmethod[window<%> set-label] when @racket[label] is a
 Otherwise, sets the bitmap label for a bitmap message.
  @bitmaplabeluseisbm[label] @|bitmapiforiglabel|
 
-}}
+}
+
+@defmethod[(set-label-color [color color%]) void?]{
```
sorawee, line 79: This should be `(is-a?/c color%)`, right?

gui-doc/scribblings/gui/message-class.scrbl
```diff
@@ -10,12 +10,13 @@ A message control is a static line of text or a static bitmap. The
...
 
-@defconstructor[([label (or/c label-string? (is-a?/c bitmap%) 
+@defconstructor[([label (or/c label-string? (is-a?/c bitmap%)
                               (or/c 'app 'caution 'stop))]
-                 [parent (or/c (is-a?/c frame%) (is-a?/c dialog%) 
+                 [parent (or/c (is-a?/c frame%) (is-a?/c dialog%)
                                (is-a?/c panel%) (is-a?/c pane%))]
                  [style (listof (or/c 'deleted)) null]
                  [font (is-a?/c font%) normal-control-font]
+                 [color (is-a?/c color%) _unset]
```
sorawee, line 19: Would it make sense to make this `(or/c #f (is-a?/c color%))` and default to `#f`?

gui-doc/scribblings/gui/message-class.scrbl
Bogdanp: Whoops, yes, thanks!

mflatt:
This looks great! Two things I noticed when I tried it:

* The initialization argument and color-setting methods should probably accept a string and convert it with `(send the-color-database find-color ....)`.

* The initialization argument is `color`, but the methods are `...label-color`. It would be better to make those the same, and I'd be inclined to go with `...color` instead of making the initialization argument `label-color`.

Bogdanp:
Thanks! I've made those changes.

mflatt:
Thanks! Ok if I squash while merging?

Bogdanp:
Yup!

PR #243, opened 2021-09-22
Add 'new-button style to tab-panel%

jjsimpso:
The 'new-button style adds a '+' button to the right of the last tab in the panel to support inserting new tabs to the panel. This implementation is for the platform-independent wxtabcanvas only.

My app which uses this code is here: https://github.com/jjsimpso/molasses/tree/tab-panel-experiment

### DIFF
--- a/gui-doc/scribblings/gui/tab-panel-class.scrbl
+++ b/gui-doc/scribblings/gui/tab-panel-class.scrbl
@@ -24,8 +24,8 @@ The @racket[tab-panel%] class does not implement the virtual
                             . -> . any) 
                            (lambda (b e) (void))]
                  [style (listof (or/c 'no-border
-                                      'can-reorder 'can-close 'flat-portable
-                                      'deleted))
+                                      'can-reorder 'can-close 'new-button
+				      'flat-portable 'deleted))
                         null]
                  [font (is-a?/c font%) normal-control-font]
                  [enabled any/c #t]
@@ -69,6 +69,11 @@ If the @racket[style] list includes @racket['no-border], no border is
  the @racket['flat-portable] flag is
  effectively always included in @racket[style] on Windows if either
  @racket['can-reorder] or @racket['can-close] is included.
+ If the @racket[style] list includes @racket['new-button] and the
+ platform-independent implementation is used for the tab control,
+ then a new tab button is added to the right of the last tab to allow
+ inserting new tabs. If the new tab button is clicked,
+ @method[tab-panel% on-new-request] is called.
  @DeletedStyleNote[@racket[style] @racket[parent]]{tab panel}
 
 @FontKWs[@racket[font]] @WindowKWs[@racket[enabled]] @SubareaKWs[] @AreaKWs[]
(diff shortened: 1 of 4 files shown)


rfindler:
Thank you for this!

When I make many entries (enough to generate the scroll bars), then the button seems to be drawn past the end (so it isn't seen).

Also, more minor: I think the button is a bit too far to the right.

jjsimpso:
You're welcome!

> When I make many entries (enough to generate the scroll bars), then the button seems to be drawn past the end (so it isn't seen).

I will take a look at this. What behavior would you prefer instead? Would it be better to always draw the new button or should we just allow it to be reached by scrolling far enough? If we always draw the button, should it be to the left or to the right of the scroll right button? If you don't have a clear preference, I can experiment.

> Also, more minor: I think the button is a bit too far to the right.

I think I agree. I believe I just used the already existing horizontal margin, but maybe I should divide that by half.

rfindler:
I would probably try something where one scrolls over to find the button
(so it would be inside the scroll bars) first. That sounds best to me but
if your experiments suggest otherwise I am sure you can convince me.
Thanks!

jjsimpso:
It turned out to be an easy fix to make the new button reachable via the scroll bars. I looked into doing what firefox does -- put the new button to the right of the scrollable region -- but that was going to be quite a bit more complicated. I think this works well enough.

I also adjusted the margin around the new button.

rfindler:
This is great, thanks. I tried another thing: the clicking behavior has to be a bit more complicated. I think it should look like the close button when you click and move off and on the button (that is, when you move the mouse away from the button but don't release the mouse button).

jjsimpso:
Ah, I see. I hadn't noticed that. I will take a look at that as well.

rfindler:
Sadly, there are supposed to be four different states:
- the mouse is not over the button, 
- the mouse is over the button but not pushed, 
- the mouse is over the button and it is pushed, 
- the mouse is not over the button, but the last time it was pushed, it was over the button

It looks like the last state is missing. Relatedly, if the button is released in that last state, the callback should not be invoked (but it should be if the button is released in the third state).

And as I play around with the close buttons, I see that there is a missing refresh for when the mouse button is pushed. If you push the button and then wiggle the mouse a little (so it stays in the circle of the close button), it'll change color. That is the color it is supposed to be, immediately after pushing the button.

jjsimpso:
OK, hopefully I got it this time. I think the four states for the new and close buttons are working correctly now. All the button mouse over states were being reset when clicked, so I fixed that and added a fourth state for the new button.

For the new button there is now a different color for each state. I can switch the colors around if you'd like. I'm reusing previously defined colors to keep things similar to the close button.

rfindler:
I'm not spotting any problems with the tab panel now! I'm ready to merge, but probably these commits should be squashed into one commit -- oh, and let's bump the version number so people can depend on this change if they need to (so update the number in the history annotation too).

If you want I can do that.

Thank you for this!

gui-doc/scribblings/gui/tab-panel-class.scrbl
```diff
@@ -24,8 +24,8 @@ The @racket[tab-panel%] class does not implement the virtual
                             . -> . any) 
                            (lambda (b e) (void))]
                  [style (listof (or/c 'no-border
-                                      'can-reorder 'can-close 'flat-portable
-                                      'deleted))
+                                      'can-reorder 'can-close 'new-button
+				      'flat-portable 'deleted))
```
sorawee, line 28: Can this be aligned with `'can-reorder` above?

jjsimpso:
Great! I used a new version in my history annotation in anticipation of the version being bumped. Just now I bumped the version in gui-lib to 1.62 to match. Did I do that correctly and is that all I need to do?

I also fixed the whitespace issue in the doc.

rfindler:
Thanks! I've rebased and pushed to master.

PR #246, opened 2021-10-03
cocoa: allow duplicate choice% labels

Bogdanp:
`NSPopUpButton` checks for duplicates before inserting new items, which
means code like

    (new choice% [choices '("a" "a" "b")] ...)

would previously panic with

    "Invalid parameter not satisfying: index <= [_item Array count]"

because the implementation would attempt to insert the "b"
label at position 2, thinking two "a" labels were inserted when only
one was.

This change uses the underlying NSMenu methods to append items since
they don't check for duplicates.  In addition to fixing the crash, it
brings the implementation in line with the one on Linux, where
duplicates are allowed.  I assume the Windows implementation also
allows duplicates, but I haven't tried it.

### DIFF
--- a/gui-lib/mred/private/wx/cocoa/choice.rkt
+++ b/gui-lib/mred/private/wx/cocoa/choice.rkt
@@ -30,18 +30,21 @@
   (inherit get-cocoa init-font register-as-child)
 
   (super-new [parent parent]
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #269, opened 2022-03-30
scaling: work around an upstream bug on wayland

capfredf:
`gdk_screen_get_monitor_geometry` is supposed to return the logic size of a
display. However, due to the upstream bug [1], we have to manually recalculate
the size.

[1] https://gitlab.gnome.org/GNOME/gtk/-/issues/2599

### DIFF
--- a/gui-lib/mred/private/wx/gtk/frame.rkt
+++ b/gui-lib/mred/private/wx/gtk/frame.rkt
@@ -16,15 +16,15 @@
          "widget.rkt"
          "cursor.rkt"
          "pixbuf.rkt"
-	 "resolution.rkt"
-	 "queue.rkt"
+         "resolution.rkt"
+         "queue.rkt"
          "../common/queue.rkt")
 
 (provide 
  (protect-out frame%
               display-origin
               display-size
-	      display-count
+              display-count
               display-bitmap-resolution
               location->window
               get-current-mouse-state
@@ -95,6 +95,10 @@
   #:make-fail make-not-available)
 (define-gtk gtk_widget_get_allocated_height (_fun _GtkWidget -> _int)
   #:make-fail make-not-available)
+
+(define-gtk gtk_get_major_version (_fun -> _uint) #:fail (lambda () (lambda () 2)))
+(define-gtk gtk_get_minor_version (_fun -> _uint) #:fail (lambda () (lambda () 2)))
+(define-gtk gtk_get_micro_version (_fun -> _uint) #:fail (lambda () (lambda () 2)))
 
 (define-gtk gtk_layout_new (_fun (_pointer = #f) (_pointer = #f) -> _GtkWidget))
 (define-gtk gtk_layout_put (_fun _GtkWidget _GtkWidget _int _int -> _void))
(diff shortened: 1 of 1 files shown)


capfredf:
I ran into this problem when I was running slideshow on a 4k display with a scale@2x on wayland. 
The resulting slides were stretched:

mflatt:
I'm not sure I follow, but it looks like this will produce the right answer with GTK3 >= 3.24.9 and the wrong answer for earlier versions (and perhaps a future repaired version). Is that true, or am I misreading?

capfredf:
Yes, that is correct. If the upstream issue is fixed, the patched function will produce a wrong answer.  As to earlier versions of gtk, we can only apply this patch only with gtk3 >= 3.24.9, but I don't know what is an idiomatic way to do that in `racket/gui`.

capfredf:
I tried `gdk_monitor_get_geometry` when I was working this patch. It produced the same incorrect answer.

capfredf:
`gtk does provide constants GTK_MAJOR_VERSION, GTK_MICRO_VERSION and GTK_MINOR_VERSION`
I took another look. Gtk also provides three functions to get there version numbers.

mflatt:
Just to make sure, does `gdk_monitor_get_workarea` have the same scaling issue?

capfredf:
It produces the same result

```
(define-gdk gdk_monitor_get_workarea (_fun _GdkMonitor _GdkRectangle-pointer -> _void))
(define-gdk gdk_display_get_default (_fun -> _GdkDisplay))
(define-gdk gdk_display_get_monitor (_fun _GdkDisplay _int -> _GdkMonitor))

(let* ([d (gdk_display_get_default)]
       [m (gdk_display_get_monitor d 0)]
       [r (make-GdkRectangle 0 0 0 0)])
  (define scale (gdk_screen_get_monitor_scale_factor
                 (gdk_screen_get_default)
                 0))
  (gdk_monitor_get_workarea m r)
  (eprintf "scale ~a x ~a y ~a width ~a height ~a~n"
           scale
           (GdkRectangle-x r)
           (GdkRectangle-y r)
           (GdkRectangle-width r)
           (GdkRectangle-height r)))
```
output:
```
scale 2 x 3840 y 0 width 3840 height 2160
```

gui-lib/mred/private/wx/gtk/frame.rkt
```diff
@@ -595,7 +599,20 @@
     (unless (num . < . (gdk_screen_get_n_monitors s))
       (fail))
     (gdk_screen_get_monitor_geometry s num r)
-    r))
+    (cond
+      ;; work around the upstream issue https://gitlab.gnome.org/GNOME/gtk/-/issues/2599
+      [(and wayland?
+            (= (gtk_get_major_version) 3)
+            (>= (gtk_get_minor_version) 24)
+            (>= (gtk_get_micro_version) 9))
```
mflatt, line 607: I think this `and` probably doesn't do what you intend for a minor version that's strictly more than 24.

mflatt:
I never found a better strategy than checking the version, so I'm ok with this approach.

Besides the correction to the version comparison, it appears that these function are not available from Gtk 2. A combination of using `#:fail` in `define-gtk` and using `gtk3?` in the conditional is probably the right idea.

PR #270, opened 2022-05-15
wxtabcanvas: fix painting with no items

Bogdanp:
This fixes a division by zero error when the `flat-portable` style is used with no tabs (which the contract for `tab-panel%` allows and works with at least the native style on macOS).

Failing example:

```racket
#lang racket/gui

(define f (new frame%
               [label "Test"]))
(define t (new (class tab-panel%
                 (super-new)
                 (define/override (on-new-request)
                   (send this set '("a"))))
               [parent f]
               [choices '()]
               [style '(no-border flat-portable new-button)]))
(send f show #t)
```


mflatt:
Thanks!

PR #277, opened 2022-07-14
Add grapheme support to the editor classes

mflatt:
Builds on racket/racket#4350, racket/draw#49, and racket/snip#9.

Changing `text%` so that positions correspond to glyphs is the obvious choice, but that change appears to be unworkable. There are too many things in the current API and uses that make sense only when position delta = character count. In the approach here, graphing counting is added separately, alongside line counting and graphical-size counting.

The `snip%` class gets three new methods: `get-grapheme-count`, `grapheme-position` (converts from graphemes to characters), and`position-grapheme` (covert from characters to graphemes). The default implementares the same as `get-count` and identity functions, respectively, but they're specialized for `string-snip%`.
    
The `text%` class also gets `grapheme-position` and `position-grapheme`, plus some grapheme sensitively (automatic for characters, on demand for strings) in `insert` and (automatic) in `move-position`, `delete`, and `on-default-char`. For compatibility, however, items and positions still corerspond to characters.

External extensions of an editor, such as keybindings would need to be adjusted to work right, but that can be done incrementally. Meanwhile, they won’t be any more broken from a user's perspective than now.

### DIFF
--- a/gui-doc/scribblings/gui/editor-overview.scrbl
+++ b/gui-doc/scribblings/gui/editor-overview.scrbl
@@ -225,6 +225,40 @@ When an editor is drawn into a display, each snip and position has a
  editor. Locations in an editor are only meaningful when the editor is
  displayed.
 
+@subsection[#:tag "graphemes"]{Characters, Graphemes, and Source Locations}
+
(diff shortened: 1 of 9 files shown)


mflatt:
Merged

PR #279, opened 2022-08-22
revised grapheme approach: grapheme positions by default

mflatt:
Commit ee68056a75 added support for graphemes and grapheme positions, but left all existing methods as using char-based positions. That approach was based on the fact that many things are deeply tied to characters instead of graphemes, especially internally, and so isn't practical to make grapheme counting the only interface.

This commit shifts to grapheme counting by default, which is enabled by keeping the old methods with a `/char` suffix while adding new versions of the old methods to perform grapheme<->char converions on entry and exit. This is not a backward-compatible change, because it shifts the relationship of existing methods and characters. It's a more forward-compatible: it makes most existing code work the right way with multi-character graphemes, and implicitly adapts an interface like `color-textoid<%>` to support multi-character graphemes. For the rare case that this change breaks a program, the program could be changed to use `/char` methods.

The `snip%` interface is more "internal" than "external", and it sticks a char-based interface, continuing to rely on explicit support for grapheme<->char conversion. The associated snip changes are https://github.com/mflatt/snip/tree/grapheme2.


mflatt:
The intent is that `grapheme-position` and `position-grapheme` from ee68056a759732682a4d9224ed3cfe249cbc382a go away, but they're still here as no-ops to ease the transition.

Meanwhile, at least once change is needed to adapt `drracket`: https://github.com/mflatt/drracket/commit/434d55d9eeec90862df9f7ccfdabb86d12eb6b0b.

mflatt:
Merged, but also added `use-char-as-grapheme` to enable the old behavior.

PR #282, opened 2022-09-19
gtk: round down scaled arguments to make-GdkRectangle

### LINKED ISSUE #280: DrRacket launch error in Ubuntu 22.04
I am running Ubuntu 22.04, launching DrRacket 8.6 (or 8.2).  I have installed them at different times using "sudo apt install racket" and also "sudo snap install racket", and in both cases, when trying to launch DrRacket from the command line, I get the following error.

```ptr-set!: given value does not fit primitive C type
  C type: _int32
  value: 963/2
  context...:
   .../ffi/unsafe.rkt:1917:18: make-GdkRectangle
   /usr/racket/share/pkgs/gui-lib/mred/private/wx/gtk/frame.rkt:620:0: display-origin
   /usr/racket/share/pkgs/gui-lib/mred/private/wxtop.rkt:53:6
   body of "/usr/racket/share/pkgs/gui-lib/mred/private/wxtop.rkt"
```

capfredf:
closes #280 and fixes the ci issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        vm: ['regular', 'CS']
+        vm: ['BC', 'CS']
 
     steps:
     - uses: actions/checkout@master
-    - uses: Bogdanp/setup-racket@v0.10
+    - uses: Bogdanp/setup-racket@v1.8.1
       with:
         architecture: 'x64'
         distribution: 'minimal'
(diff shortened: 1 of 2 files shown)


capfredf:
@samth can you take a look?

samth:
Should this use `inexact->exact` as well? Does `_int` require an exact integer?

samth:
I'm ok with merging this; I don't know if @mflatt has any thoughts.

mflatt:
LGTM

PR #283, opened 2022-09-23
Cancel writing events after the Kill button is pressed

sorawee:
When we press the kill button, make it so that queued up "do-insertion"s are not processed.

This code is originally authored by @rfindler

Fixes racket/racket#1677 and racket/drracket#444

### DIFF
--- a/gui-lib/framework/private/text-port.rkt
+++ b/gui-lib/framework/private/text-port.rkt
@@ -290,7 +290,10 @@
         (channel-put box-read-chan (cons eof (position->line-col-pos unread-start-point))))
       (define/public-final (clear-input-port) (channel-put clear-input-chan (void)))
       (define/public-final (clear-box-input-port) (channel-put box-clear-input-chan (void)))
-      (define/public-final (clear-output-ports) 
+
+      (define output-ports-clear-count 0)
+      (define/public-final (clear-output-ports)
+        (set! output-ports-clear-count (add1 output-ports-clear-count))
         (channel-put clear-output-chan (void))
         (init-output-ports))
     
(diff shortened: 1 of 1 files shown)


gui-lib/framework/private/text-port.rkt
```diff
@@ -290,7 +290,10 @@
         (channel-put box-read-chan (cons eof (position->line-col-pos unread-start-point))))
       (define/public-final (clear-input-port) (channel-put clear-input-chan (void)))
       (define/public-final (clear-box-input-port) (channel-put box-clear-input-chan (void)))
-      (define/public-final (clear-output-ports) 
+
+      (define output-ports-clear-count 0)
+      (define/public-final (clear-output-ports)
+        (set! output-ports-clear-count (add1 output-ports-clear-count))
```
sorawee, line 296: @rfindler this is the only place that writes to `output-ports-clear-count`, so provided that there is only one thread that calls `(clear-output-ports)`, I think there won't be a race condition?

capfredf:
I've fixed the CI issue in #282

rfindler:
There's a race because the `set!` happens in one thread and the lookup (`(define current-clear-count output-ports-clear-count)`) happens in a different thread. I think that because racket guarantees sequential consistency for `thread` (it doesn't for futures or places) that this race condition is benign and probably this is okay, overall.

That said, I feel like it would be better if there were some way to use the existing syncronization that's happening in this code to be able to cancel the pending work.

sorawee:
Closing in favor of https://github.com/racket/gui/commit/b51090c0a14109cb30538cc7752c23c6a2186b4e

PR #292, opened 2023-03-22
add application-dark-mode-handler

rfindler:
Questions I wonder about:

- did I do something dumb at the FFI level?
- Is `application-dark-mode-handler` a good name?
- should we add also a `dark-mode?` function (that would just do the same thing as [`mrlib/panel-wob`'s `white-on-black-panel-scheme?`](https://docs.racket-lang.org/mrlib/White_on_Black_Panel_Predicate.html)?

If not for the last, the docs should be adjusted to point to `mrlib/panel-wob`.


mflatt:
This all looks fine to me. I have no opinion whether `dark-mode?` is needed.

For what it's worth, I would have put `observeValueForKeyPath` in `RacketApplicationDelegate` and used `app-delegate` (or `self` within `applicationDidFinishLaunching:`) as the observer object, instead of allocating a new class and observer object. An advantage of your approach is that the `observeValueForKeyPath` method doesn't have to check which key it's called for, in case we need to observe more keys in the future.

PR #293, opened 2023-03-23
refreshing canvases when entering/leaving dark mode

rfindler:
There is at least one problem with this pr, namely that it only requests refreshes for the frames in the original/main eventspace. I had thought of using the custodian hierarchy to reach other eventspaces but this doesn't work because I need a parent custodian that I don't have.

I'm also not sure if `queue-paint` is the right method to invoke when finding a canvas.

### DIFF
--- a/gui-lib/mred/private/wx/cocoa/frame.rkt
+++ b/gui-lib/mred/private/wx/cocoa/frame.rkt
@@ -48,6 +48,18 @@
 ;;  problems.
 (define all-windows (make-hash))
 
+(set-queue-events-to-refresh-all-canvases!
+ (let ([queue-events-to-refresh-all-canvases
+        (λ ()
+          (atomically
+           (for ([b (in-hash-values all-windows)])
+             (define frame (weak-box-value b))
+             (when frame
+               (queue-event
+                (send frame get-eventspace)
+                (λ () (send frame request-refresh-all-canvas-children)))))))])
+   queue-events-to-refresh-all-canvases))
+
 ;; called in atomic mode
 (define (send-screen-change-notifications flags)
   (reset-menu-bar!)
--- a/gui-lib/mred/private/wx/cocoa/queue.rkt
+++ b/gui-lib/mred/private/wx/cocoa/queue.rkt
@@ -30,7 +30,8 @@
               try-to-sync-refresh
               try-to-flush
               sync-cocoa-events
-              set-screen-changed-callback!)
+              set-screen-changed-callback!
+              set-queue-events-to-refresh-all-canvases!)
 
  ;; from common/queue:
  current-eventspace
@@ -102,6 +103,7 @@
       ;; we add an observer only for "effectiveAppearance",
       ;; so no check is needed to know why we got here
       (queue-dark-mode-event)
+      (queue-events-to-refresh-all-canvases)
       (void)]
   [-a _BOOL (applicationShouldHandleReopen: [_id app] hasVisibleWindows: [_BOOL has-visible?])
       ;; If we have any visible windows, return #t to do the default thing.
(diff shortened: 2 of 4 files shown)


LiberalArtist:
IIUC, this PR and https://github.com/racket/gui/pull/292/files detect changes to the system "dark mode" state only on Mac, but I think it should be possible to support across platforms. I know at least that many apps react properly for me on KDE Plasma; I think it also works on Gnome and Windows (11?).

I'm not sure what the best way is to listen for a change on various platforms.

With Gtk, it seems like it might be enough to listen for changes to [`gtk-theme-name`](https://docs.gtk.org/gtk3/property.Settings.gtk-theme-name.html).

On Windows, the de facto way seems to be using [`RegNotifyChangeKeyValue`](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regnotifychangekeyvalue) to detect changes to the `AppsUseLightTheme` value of `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Themes\Personalize` (which may not exist).

rfindler:
These are both fixing problems that are mac os specific. I'm guessing there are similar problems on other platforms but probably they are different and require different fixes. If you're interested in improving dark mode on gtk and windows! I've looked into windows (and there is an issue discussing it somewhere, I think) and the OS doesn't seem to provide a way to detect the theme. But Gtk probably just needs some love.

rfindler:
(Oh, sorry, #292 isn't macos specific -- filling in something that calls the dark-mode handler on other platforms would help them for sure!)

LiberalArtist:
> These are both fixing problems that are mac os specific. I'm guessing there are similar problems on other platforms but probably they are different and require different fixes.
> [&hellip;]
> (Oh, sorry, https://github.com/racket/gui/pull/292 isn't macos specific -- filling in something that calls the dark-mode handler on other platforms would help them for sure!)

That's mostly what I had in mind. (But I didn't understand that there was a Mac-specific problem involved.)

> If you're interested in improving dark mode on gtk and windows that would be great! I've looked into windows (and there is an issue discussing it somewhere, I think) and the OS doesn't seem to provide a way to detect the theme.

That's my understanding, too, as far as detecting the theme. There's a [Microsoft article](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-windows-themes#know-when-dark-mode-is-enabled) recommending basically the same strategy [`white-on-black-panel-scheme?`](https://docs.racket-lang.org/mrlib/White_on_Black_Panel_Predicate.html) uses, but using a Windows-specific API. The registry key I mentioned is undocumented, but seems to be used by a number of projects, like Chromium, to detect *changes* to the theme.

> But Gtk probably just needs some love.

I hope to supply some more broadly (in the amorphous medium term).

rfindler:
I'm working on other changes that aren't platform specific that are also helping, like [this one](https://github.com/racket/drracket/commit/e28d0b893ba3684a17a9223bc97a550eed0c15bf) and I just found a source of a lot of flicker, so doing the os-specific pieces will be even more impactful.

mflatt:
To get all frames, each backend could export a function that can be used to get that list. For Mac OS, there's an `all-windows` table defined in `frame.rkt`. For GTK, it's `all-frames` in `frame.rkt`. For Windows, there doesn't seem to be a table right now, but `all-frames` could be added.

When you can get all frames, then for each frame `f`, `(queue-window-event f (lambda () (send f request-refresh-all-canvas-children)))` is probably the right way to queue the request (in the right eventspace).

rfindler:
I took a look into adding `get-all-frames` functionality to all of the backends but ran into cyclic dependencies (specifically the function that iterates over all the frames to call the `request-refresh-all-canvas-children` method seems to go in common/queue.rkt, but this library seems to actually be something that each platform might use -- not that can use the platforms). 

So, instead, I just stuck with putting it into the cocoa backend for now since, as far as we know, that's the only backend that needs to use this functionality. I do see that when I switch from light to dark mode under ubutntu 22, that the run button, etc. on the drracket frame don't change color, but I also see that mousing over them also doesn't cause them to change color so it isn't clear what needs to be done there, at least not yet. 

I've updated the PR so that there is still a method on all frames that refreshes everything but it is only called on cocoa and we only get the list of all the frames (for this purpose) on cocoa.

rfindler:
PS: I think the github actions fail because they need bug fixes in `snip-lib` but we don't have a snapshot build with that in it yet.

mflatt:
LGTM

rfindler:
Thanks!

PR #294, opened 2023-03-26
Update gtk to react to dark mode changes

rfindler:
Things I wonder about:

- Is the deallocation of the value in `the-text-style` right?

- Is it okay to add one more set of declarations for `_GtkSettings` and related things (in style.rkt)?

- Should there be some versioning guards somewhere in this code?

And, of course, using a regular expression on the name of the theme also seems somewhat sketchy but I'm not able to turn up something better.

### DIFF
--- a/gui-lib/mred/private/wx/gtk/style.rkt
+++ b/gui-lib/mred/private/wx/gtk/style.rkt
@@ -2,7 +2,9 @@
 (require ffi/unsafe
          "types.rkt"
          "utils.rkt"
-         "init.rkt")
+         "init.rkt"
+         "frame.rkt"
+         "../common/queue.rkt")
 
 (provide 
  (protect-out get-selected-text-color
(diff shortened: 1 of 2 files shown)


rfindler:
This seems to work on ubuntu 22 (I didn't try it on any other versions)

gui-lib/mred/private/wx/gtk/style.rkt
```diff
@@ -96,3 +101,31 @@
...
   (extract-color-values (GtkStyle-bg1 the-text-style)))
+
+(define _GtkSettings (_cpointer 'GtkSettings))
+(define-gtk gtk_settings_get_default (_fun -> _GtkSettings))
+(define-gtk g_object_notify (_fun _GtkSettings _string -> _void))
+(define-gobj g_object_get/string (_fun _GtkSettings _string (r : (_ptr o _pointer)) (_pointer = #f)
+                                       -> _void
+                                       -> r)
+  #:c-id g_object_get)
+(g_object_notify (gtk_settings_get_default) "gtk-theme-name")
```
mflatt, line 112: Is this needed? I'd expect `g_object_notify` to send a notification, while the intent here generally is to receive notifications.

mflatt:
The deallocation and new declarations look fine, and it appears only only old functionality is used (i.e., no version guards needed). Aside from whether `g_object_notify` needs to be called, this looks good to me.

gui-lib/mred/private/wx/gtk/style.rkt
rfindler: Yep, as expected, removing those two lines works well.

rfindler:
Thanks!

PR #296, opened 2023-04-21
adjust the byte conversion loop in open-output-text-editor to pull ou…

rfindler:
…t the bytes that are part of `'aborted` results from the byte coverter and save them for later outside of the pipe, instead of inside of the pipe

### DIFF
--- a/gui-lib/mred/private/snipfile.rkt
+++ b/gui-lib/mred/private/snipfile.rkt
@@ -354,7 +354,7 @@
       (define (show s)
         (define (insert)
           (send text begin-edit-sequence)
-          (send text insert s pos)
+          (send text insert s pos 'same #t #t)
           (send text end-edit-sequence))
         (if (and eventspace
                  (and (not (eq? (current-thread) 
--- a/gui-lib/mred/private/snipfile.rkt
+++ b/gui-lib/mred/private/snipfile.rkt
@@ -364,12 +364,12 @@
             (insert))
 	(set! pos (+ (string-length s) pos)))
       (define (flush-text)
-	(let ([cnt (peek-bytes-avail!* raw-buffer 0 #f in)])
+	(let ([cnt (peek-bytes-avail!** raw-buffer 0 #f in)])
 	  (when (positive? cnt)
 	    (let-values ([(got used status) (bytes-convert cvt raw-buffer 0 cnt utf8-buffer)])
 	      (cond
 	       [(positive? got) 
-		(read-bytes-avail!* raw-buffer in 0 used)
+                (read-bytes! raw-buffer in 0 used)
 		(show (bytes->string/utf-8 utf8-buffer #\? 0 got))
 		(flush-text)]
 	       [(eq? status 'error) 
@@ -412,4 +412,11 @@
 		     (add1 pos))))
 	 void
 	 (add1 pos)))
-      port)))
+      port))
+
+  (define (peek-bytes-avail!** bstr skip progress-evt in)
+    (let loop ([got 0])
+      (define cnt (peek-bytes-avail!* bstr (+ skip got) progress-evt in got))
+      (if (zero? cnt)
+          got
+          (loop (+ got cnt))))))
--- a/gui-lib/mred/private/wxme/text.rkt
+++ b/gui-lib/mred/private/wxme/text.rkt
@@ -1923,14 +1923,16 @@
       (set! insert-force-streak? ifs?)
       (set! typing-streak? #t)))
 
-  (define/private (do-insert-graphemes str start end scroll-ok?)
+  (define/private (do-insert-graphemes str start-in end-in scroll-ok?)
     ;; maybe join characters to form a grapheme; we limit
     ;; the search for a grapheme to one surrounding snip on
     ;; the grounds that this makes sense when graphemes are
     ;; already joined
+    (define start (max 0 (min start-in len)))
+    (define end (if (eq? end-in 'same) start (max start (min end-in len))))
     (define keep-caret? (and (= start startpos)
                              (or (eq? end 'same) (eqv? end start))))
-    (let loop ([s str] [start start] [end (if (eq? end 'same) start (max start end))])
+    (let loop ([s str] [start start] [end end])
       (define pre-s (do-find-snip start 'before))
       (define pre-pos (get-snip-position pre-s))
       (define pre-count (snip->count pre-s))
--- a/gui-test/tests/gracket/editor.rktl
+++ b/gui-test/tests/gracket/editor.rktl
@@ -394,6 +394,26 @@
   (close-output-port p)
   (test "a· ·" 'unicode-code-point-broken-up (send t get-text)))
 
+(let ()
+  (define t (new text%))
+  (send t set-styles-sticky #f)
+  (define bts #"\360\237\217\264\342\200\215\342\230\240\357\270\217")
+  (define p (open-output-text-editor t))
+  (write-bytes bts p)
+  (close-output-port p)
+  (test 1 'pirate-flag-all-at-once (send t last-position)))
+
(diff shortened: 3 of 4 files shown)


rfindler:
I'm not sure if this is the right fix but it does seem to fix the second new test case.

mflatt:
This change makes sense to me.

I wonder whether the `(send text insert ...)` in `show?` should pass enough arguments to include `#t` for `join-graphemes?`. That would be aimed at something like UTF-8 encoding parts broken across writes, but for grapheme clusters broken across characters.

rfindler:
Do you know why the version without the fix doesn't work? I couldn't figure that out.

rfindler:
This program seems to turn things into a pirate flag already; I'm not sure why!

```
#lang racket/gui
(define bts #"\360\237\217\264\342\200\215\342\230\240\357\270\217")
(define f (new frame% [width 300] [height 300] [x 1000] [label ""]))
(define txt (new text%))
(define ec (new editor-canvas% [parent f] [editor txt]))
(send f show #t)
(thread
 (λ ()
   (define p (open-output-text-editor txt))
   (for ([b (in-bytes bts)])
     (write-byte b p)
     (sleep 1/2)
     (flush-output p))
   (close-output-port p)))
```

When I change the `insert` call in `show` to this:

```
(send text insert s pos 'same #t #t)
```

I get these errors with that program:

```
string-ref: index is out of range for empty string
  index: 0
  string: ""
  context...:
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wxme/text.rkt:1926:2: do-insert-graphemes method in text%
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/snipfile.rkt:356:8: insert
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:436:6
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:487:32
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:639:3
string-ref: index is out of range for empty string
  index: 1
  string: ""
  context...:
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wxme/text.rkt:1926:2: do-insert-graphemes method in text%
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/snipfile.rkt:356:8: insert
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:436:6
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:487:32
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:639:3
string-ref: index is out of range for empty string
  index: 2
  string: ""
  context...:
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wxme/text.rkt:1926:2: do-insert-graphemes method in text%
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/snipfile.rkt:356:8: insert
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:436:6
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:487:32
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:639:3
string-ref: index is out of range for empty string
  index: 3
  string: ""
  context...:
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wxme/text.rkt:1926:2: do-insert-graphemes method in text%
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/snipfile.rkt:356:8: insert
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:436:6
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:487:32
   /Users/robby/git/plt/extra-pkgs/gui/gui-lib/mred/private/wx/common/queue.rkt:639:3
```

mflatt:
Oh, I see what you mean about why the old code looks like it should work.

I think the issue is actually that the call to `peek-bytes-avail!*` is expecting that as many bytes will be read as available, but the current implementation of `peek-bytes-avail!*` might stop early, as long as it can return at least one byte. Replacing `peek-bytes-avail!*` with `peek-bytes-avail!**` (to keep reading more as long as something is available) gets closer:

```
        (define (peek-bytes-avail!** bstr skip progress-evt in)
          (let loop ([got 0])
            (let ([cnt (peek-bytes-avail!* raw-buffer (+ skip got) progress-evt in got)])
              (if (zero? cnt)
                  got
                  (loop (+ got cnt))))))
```

The subsequent call to `read-bytes-avail!` seems like it has the same problem and should be just `read-bytes!`.

Changing those two calls makes the example work for me, and I think it's probably more on the right track. The candidate repair in this PR doesn't cooerate with `force-text` as intended, I think.

rfindler:
Using `peek-bytes-avail!**` seems to make all the test suites pass.

rfindler:
I've made a little bit of progress on the strange blinking problem. First piece of information is that it doesn't require `open-output-text-editor` at all; this program reproduces it:

```
#lang racket/gui

(define text (new text%))
(send text insert "az")
(send text insert (bytes->string/utf-8 #"\360\237\217\264") 1 'same #t #t)
(send text insert "\u200D" 2 'same #t #t)
(send text change-style (make-object style-delta% 'change-bold) 1 2)
(send text insert (bytes->string/utf-8 #"\342\230\240") 3 'same #t #t)
(send text insert (bytes->string/utf-8 #"\357\270\217") 4 'same #t #t)

(define f (new frame% [label ""] [width 400] [height 400]))
(define ec (new editor-canvas% [parent f] [editor text]))
(send f show #t)
```

Second, it looks like the code is organized such that the function [`process-snips`](https://github.com/racket/gui/blob/master/gui-lib/mred/private/wxme/text.rkt#L5595) returns the two "x" coordinates, `hsxs` and `hsxe`. If they are the same, then we draw a vertical line and if they are different we're going to draw a highlight-color rectangle. Furthermore, it looks like `hsxs` is wrong -- it is returning the x value for after the "z" whereas `hsxe` is (correctly) the one for after the "z". 

... after that, I got a bit lost.

mflatt:
For the blinking bug, the problem was in `string-snip%`, addressed by racket/snip@863b13662b5e6ae6271c33a41969cac4f57a4ef6. Thanks to @rfindler  for adding the test here, since tests for snips are still in this repo.

mflatt:
I pushed a commit to make two tests check the grapheme count, instead of the character count, since the intent is to make sure that graphemes are joined.

mflatt:
The test was failing becuase it used `(= i 1)` where `(= i 0)` was intended. That is, the style should be changed for the first character, and then joining correctly will will combine with that styled character. But setting a middle character's style alone will explicitly break the grapheme.

An earlier variant of the test was meant specifically for a text output port, so I added that variant.

rfindler:
> The test was failing becuase it used `(= i 1)` where `(= i 0)` was intended. That is, the style should be changed for the first character, and then joining correctly will will combine with that styled character. But setting a middle character's style alone will explicitly break the grapheme.
> 
> An earlier variant of the test was meant specifically for a text output port, so I added that variant.

Oh! That totally makes sense.

I'll tidy up the commits here and push to master; I don't think there are any remaining issues.

rfindler:
Okay, I've combined the commits based on the functionality and pushed them to master. 

Thanks!

PR #297, opened 2023-06-09
tex-table: add guillemetleft and guillemetright

AlexKnauth:
guillemetleft for `«`, and guillemetright for `»`.

See also https://github.com/racket/scribble/pull/367.

When I try to run it, I get an error message like this:
guillemet-error

### DIFF
--- a/tex-table/tex-table.rkt
+++ b/tex-table/tex-table.rkt
@@ -157,6 +157,8 @@
     ("simeq" "≃")
     ("ll" "≪")
     ("gg" "≫")
+    ("guillemetleft" "«")
+    ("guillemetright" "»")
     ("asymp" "≍")
     ("parallel" "∥")
     ("subset" "⊂")


mfelleisen:
Which package did you load/how to get the two chars? (They are used in French and Russian, so one of those?)

AlexKnauth:
I get the `Undefined control sequence` error above when using `\guillemetleft` and `\guillemetright`, but when I manually edit the file to introduce the typo `guillemot` instead of `guillemet`, then it renders the pdf.

Could this be because I have an old version of LaTeX? Or an old version of one of the packages? How can I make it use a newer version that has the correct `guillemet` spelling?

mflatt:
How about using the old name in Scribble (since that works for older Latex versions) while DrRacket's input uses only the new name?

mflatt:
I think there's a good chance that we'll want things to work on older Latex installations, since our build machines tend to have older version — especially if it's just a matter of using the old names in the Scribble LaTeX backend.

rfindler:
Would it be good to use the tex command "define this if not defined" that I think we've used before but I forget how it goes and then make scribble generate the new command?

AlexKnauth:
Would a "define this if not defined" command need to go somewhere in `scribble-lib/scribble/scribble-load.tex` or something related to https://github.com/racket/scribble/pull/367 ?

mflatt:
Are we worried that `\guillemotleft` and `\guillemotright` (the old spellings) will disappear from LaTeX?

AlexKnauth:
The `guillemet` versions have been in LaTeX since 2018, so like 5 years now.

rfindler:
> Are we worried that `\guillemotleft` and `\guillemotright` (the old spellings) will disappear from LaTeX?

Oh, this is a good point! That sure seems easiest.

AlexKnauth:
I've added another commit to https://github.com/racket/scribble/pull/367 to use the old name `guillemot` in Scribble output:
https://github.com/racket/scribble/pull/367/commits/2892288de29f916ba08969c8262089c140de2775

mflatt:
Thanks!

rfindler:
@AlexKnauth it looks like these changes broke the redex documentation. It looks to me like there needs to be some delimiteres somewhere. In the generated latex for the redex docs I see:

```
\begin{SCodeFlow}\begin{RktBlk}\begin{SingleColumn}\RktPn{(}\RktSym{\badlink{\RktValLink{collection{-}file{-}path}}}\mbox{\hphantom{\Scribtexttt{x}}}\textbf{\guillemotleftfilename.rkt\guillemotright}

\mbox{\hphantom{\Scribtexttt{xxxxxxxxxxxxxxxxxxxxxx}}}\RktVal{"redex"}

\mbox{\hphantom{\Scribtexttt{xxxxxxxxxxxxxxxxxxxxxx}}}\RktVal{"examples"}

\mbox{\hphantom{\Scribtexttt{xxxxxxxxxxxxxxxxxxxxxx}}}\RktVal{"define{-}judgment{-}form"}\RktPn{)}\end{SingleColumn}\end{RktBlk}\end{SCodeFlow}

\noindent replacing \textbf{\guillemotleftfilename.rkt\guillemotright} with one of the names listed above.

```

which looks like it has "filename.rkt" right up against `\guillemotleft` but there should probably be a `{}` between them.

rfindler:
Oh wait, sorry, wrong place. I'll comment in the other place.

PR #308, opened 2023-09-14
message: handle invalid colors under cocoa and gtk

### LINKED ISSUE #307: `set-color` method in `message%` signals internal error
The program
```
#lang racket/gui
(send (new message% [parent (new frame% [label ""])] [label ""]) set-color "notacolor")
```

produces the error:

```
class-field-accessor: contract violation
  expected: object?
  given: #f
```

Bogdanp:
Changes `message%` cocoa and gtk wxs to reset messages to the default color when set to an invalid color or `#f`. The win32 implementation already behaves this way.

Fixes #307

~~Draft while I set up a Linux VM to test the change there also.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,21 +4,22 @@ on: [push, pull_request]
 
 jobs:
   build-test:
-    runs-on: ubuntu-18.04
+    name: Build & Test (Racket ${{ matrix.variant }})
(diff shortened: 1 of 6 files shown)


Bogdanp:
Bumping this so it doesn't get lost.

mflatt:
Thanks!

PR #310, opened 2023-10-22
Fix argument name in color scheme sample code

muscar:
The sample uses `#:label`, which seem to have been replaced by `#:title`.


rfindler:
Thanks!

PR #313, opened 2023-10-26
Avoid duplicate definitions of `gtk_fixed` operations.

samth:


gui-lib/mred/private/wx/gtk/group-panel.rkt
```diff
@@ -6,6 +6,7 @@
          "window.rkt"
          "client-window.rkt"
          "panel.rkt"
+         "frame.rkt"
```
mflatt, line 9: Doesn't seem like a big deal, but is this import needed?

gui-lib/mred/private/wx/gtk/tab-panel.rkt
```diff
@@ -9,6 +9,7 @@
          "types.rkt"
          "widget.rkt"
          "message.rkt"
+         "frame.rkt"
```
mflatt, line 12: This import doesn't seem needed, either.

mflatt (approved):

gui-lib/mred/private/wx/gtk/group-panel.rkt
samth: It's needed because they were getting the functions from the other definition before.

mflatt: Oh, I see. Ok!

Issue #314, opened 2023-10-28
Linux sliders are upside down


kengruven:
On Linux, a vertical slider% has the lowest value at the top, and the highest value at the bottom, and it visualy 'fills' the slider from the top down.  It gets this behavior from GTK's default.

This is backwards from how sliders work on macOS, Qt, Java, Firefox, Chrome, Safari, Edge, light switches on walls, lighting consoles, audio consoles, and every other vertical slider I've encountered.  GTK is just weird here.  Nobody pushes a slider *up* and expects a *lower* volume or *less* light.

Normally, I'm in favor of cross-platform toolkits following the local conventions, but in this case I don't see any rationale for GTK being the odd one out, and also I haven't found a single GTK application (even among the GTK demos) with a vertical slider, so I don't think there's any real convention to be beholden to.  The racket/gui documentation doesn't say which orientation you get, so we wouldn't be breaking any promises to change it.

As with most toolkits, GTK does provide a property to flip the slider vertically (though their own documentation on it is self-contradictory!), but AFAICT it's not exposed by racket/gui so programmers can't use it.

I propose that slider% on GTK set the "inverted" property on vertical sliders so the highest value is on top, as it is everywhere else in the world.  My second choice would be to expose such a property in racket/gui's slider% so that applications can choose either "platform native" or "cross-platform consistent".

sorawee:
@mflatt is this fixed? Can it be closed?

PR #315, opened 2023-10-31
add 'upward style for slider controls

mflatt:
Relevant to #314.

As far as I can tell, a vertical slider in `racket/gui` is top-to-bottom on all platforms. I don't know or remember why, but I agree that a bottom-to-top slider seems more useful, and probably non-`racket/gui` programs use bottom-to-top sliders everywhere. The top-to-bottom rendering is also inherent in the macOS control, at least when the range is >= 64.

To add bottom-to-top support without breaking any old programs — just in case — this commit adds an `'upward` style option alongside `'horizontal` and `'vertical`.

### DIFF
--- a/gui-doc/scribblings/gui/slider-class.scrbl
+++ b/gui-doc/scribblings/gui/slider-class.scrbl
@@ -23,7 +23,7 @@ Whenever the user changes the value of a slider, its callback
                                (is-a?/c panel%) (is-a?/c pane%))]
                  [callback ((is-a?/c slider%) (is-a?/c control-event%) . -> . any) (lambda (b e) (void))]
                  [init-value position-integer? min-value]
-                 [style (listof (or/c 'horizontal 'vertical 'plain 
+                 [style (listof (or/c 'horizontal 'vertical 'upward 'plain
                                       'vertical-label 'horizontal-label 
                                       'deleted)) 
                         '(horizontal)]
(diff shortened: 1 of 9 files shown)


kengruven:
It looks like GTK is the only backend which tells the native control about the direction.  Do Cocoa/Win32 have any corresponding functions?  NSSlider, for example, seems to "fill up" from the bottom (but not if there are ticks?), so a `vertical` slider could look somewhat misleading.

On win32, I see something called `TBS_DOWNISLEFT` that might be relevant, but I don't have a way to try it out.

(BTW, is there an easy way to build and test a package like racket/gui which has lots of dependencies and dependents in the stdlib?  I've looked around a little and haven't found anything yet.)

mflatt:
I couldn't find a way to get downward sliders to render right on Mac OS when the range is >= 64, so I added a note to the documentation that rendering could be misleading.

@kengruven — Thanks for the pointer to `TBS_DOWNISLEFT`, but I wasn't able to get it to do anything on Windows. Maybe I wasn't using that right somehow, but the rendering of sliders doesn't seem to suggest a direction, anyway.

For info on trying out changes to a package like `gui-lib`, see this section in "Building, Distributing, and Contributing to Racket": https://docs.racket-lang.org/racket-build-guide/contribute.html#%28part._pkg-contribute%29

PR #316, opened 2023-11-03
update the focus when deleting a snip that has the focus

rfindler:
This fixes a bug that was found [on discourse](https://racket.discourse.group/t/uncommenting-does-not-work-on-any-linux-distro-with-gnome-3/1299/6).


samth:
Can we merge this? Should this have gone into the 8.12 release?

rfindler:
I'd forgotten about this but merging it now without putting it on the release branch seems safer.

PR #317, opened 2023-11-04
Update slider-class.scrbl

kengruven:
Fix two argument names.

Also, a non-increasing range (like 0 to 0) is allowed.

### DIFF
--- a/gui-doc/scribblings/gui/slider-class.scrbl
+++ b/gui-doc/scribblings/gui/slider-class.scrbl
@@ -45,8 +45,8 @@ If @racket[label] is a string, it is used as the label for the slider.
 The @racket[min-value] and @racket[max-value] arguments specify the
  range of the slider, inclusive. The @racket[init-value] argument
  optionally specifies the slider's initial value. If the sequence
- [@racket[min-value], @racket[initial-value], @racket[maximum-value]]
- is not increasing, @|MismatchExn|.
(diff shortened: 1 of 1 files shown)


gui-doc/scribblings/gui/slider-class.scrbl
```diff
@@ -45,8 +45,8 @@ If @racket[label] is a string, it is used as the label for the slider.
 The @racket[min-value] and @racket[max-value] arguments specify the
  range of the slider, inclusive. The @racket[init-value] argument
  optionally specifies the slider's initial value. If the sequence
- [@racket[min-value], @racket[initial-value], @racket[maximum-value]]
- is not increasing, @|MismatchExn|.
+ [@racket[min-value], @racket[init-value], @racket[max-value]]
+ is not non-decreasing, @|MismatchExn|.
```
sorawee, line 49: I would avoid double negation, which is confusing to me.

My proposal is to say:

```
The sequence [@racket[min-value], @racket[init-value], @racket[max-value]] must be non-decreasing. Otherwise, @|MismatchExn|
```

kengruven: Sure, I approve of this change.  I had thought it was a little awkward at the time but I didn't have a great idea (except possibly to just write out the invariant `(<= min-value init-value max-value)`).

sorawee:
Thanks!

PR #319, opened 2023-11-15
extra symbols

bdeket:
Some symbols I miss from time to time:
double/tripple integrals
double struck uppercase letters
angle

### DIFF
--- a/tex-table/tex-table.rkt
+++ b/tex-table/tex-table.rkt
@@ -133,6 +133,7 @@
     ("land" "∧")
     ("lnot" "¬")
     ("triangleleft" "◃")
+    ("angle" "∠")
     ("odot" "⊙")
     ("star" "★")
     ("dagger" "†")
@@ -190,7 +191,11 @@
     ("coprod" "∐")
     
     ("int" "∫")
+    ("iint" "∬")
+    ("iiint" "∭")
     ("oint" "∮")
+    ("oiint" "∯")
+    ("oiiint" "∰")
 
     ("sqrt" "√")
     
@@ -198,6 +203,14 @@
     ("smiley" "☺")
     ("blacksmiley" "☻")
     ("frownie" "☹")
+
+    ("Naturals" "ℕ")
+    ("Integers" "ℤ")
+    ("Rationals" "ℚ")
+    ("Reals" "ℝ")
+    ("Complex" "ℂ")
+    ("Re" "ℜ")
+    ("Im" "ℑ")
     
     ("S" "§")
     ("l" "ł")
--- a/tex-table/tex-table.rkt
+++ b/tex-table/tex-table.rkt
@@ -204,11 +204,11 @@
     ("blacksmiley" "☻")
     ("frownie" "☹")
 
-    ("Naturals" "ℕ")
-    ("Integers" "ℤ")
-    ("Rationals" "ℚ")
-    ("Reals" "ℝ")
-    ("Complex" "ℂ")
+    ("bC" "ℂ")
+    ("bN" "ℕ")
+    ("bQ" "ℚ")
+    ("bR" "ℝ")
+    ("bZ" "ℤ")
     ("Re" "ℜ")
     ("Im" "ℑ")
     


rfindler:
Each time we add something to this list, there is a possibility that the DrRacket manual build breaks because scribble runs into new unicode characters that it isn't translating into latex properly. Probably the simplest way to find those breakages is to do this:

```
$ cd ...../drracket/drracket/scribblings/drracket
$ scribble --dest x --latex drracket.scrbl && (cd x ; pdflatex drracket)
```

where the five ellipses are the place where the drracket package is installed.

Doing that with the commit in this pull request in place results in this at the end of the latex output:

```
See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.3514 \hbox{\mbox{\hphantom{\Scribtexttt{x}}}∠
                                                 } \\
? x
```

which suggest that [this function](https://github.com/racket/scribble/blob/master/scribble-lib/scribble/latex-render.rkt#L1296) needs another line.

I'll take a look into adding these -- I just tried and `$\\angle` seems to work for the first one. (I'm out of time for the moment but I can return to this later on today, I expect.)

Another thing we might consider doing is changing the DrRacket manual but I lean towards not doing that.

rfindler:
This [wikipedia page](https://en.wikipedia.org/wiki/Blackboard_bold) looks helpful for finding the actual characters.

bdeket:
Thank you for looking into this and sorry for not following up.
Can I help out with completing this?
Do I understand correctly that `Naturals` etc. should be changed `\bN` etc.

rfindler:
I didn't merge the PR yet (but I'm happy to do so if you want!). I actually lost track of it! But the changes to Scribble are done, I think.

We can have multiple names for a given symbol so if you use a latex package that uses `\Naturals`, etc, then we can use that name. We can also use `\bN` and for any of the symbols listed in the "Unicode symbol" column of [this table on the black-board bold wikipedia page](https://en.wikipedia.org/wiki/Blackboard_bold#Usage); at least, I think those are all covered by the commit I did to scribble above.

How would you like to proceed?

tex-table/tex-table.rkt
```diff
@@ -190,14 +191,26 @@
...
+    ("oiiint" "∰")
 
     ("sqrt" "√")
     
     ("skull" "☠") 
     ("smiley" "☺")
     ("blacksmiley" "☻")
     ("frownie" "☹")
+
+    ("Naturals" "ℕ")
```
sorawee, line 207: Should it be `Naturals` or `Natural`? If former, shouldn't `Complex` below be `Complexes` for consistency?

rfindler:
My take on this is that we should use "\Naturals" if that's already
established in some fairly commonly used latex package. Ditto any other
names.

Robby

bdeket:
The standard seems to be to use the `\mathbb` approach.
I've found this list [https://www.rollpie.com/post/533](https://www.rollpie.com/post/533) which seem to come from the `texvc` package. But this lacks uniformity and misses Rationals.

sorawee:
I would vote to not follow `texvc`. It's not commonly used.

sorawee:
Wait, it looks like @rfindler already pushed the initial commit to the master branch last month.

https://github.com/racket/gui/blob/master/tex-table/tex-table.rkt#L207

sorawee:
@rfindler what do you think should be the next step? Do you think we should just cherry-pick 2d23378d7669faecc96a4f43c476e5dbc61eacde and then close the PR?

rfindler:
I can add the other \b<letter> shortcuts; they seem useful to me. Shall I do that? Maybe after cherry-picking?

sorawee:
I'd be OK with that.

rfindler:
Thanks for this @bdeket ! I've cherry-picked the commit.

shhyou:
Late to the party but +1 for the currently just-pushed `\bN`, etc. Other same abbrvs:
https://github.com/agda/agda/blob/ee876ee229cdc2f3b1ff80f6d60568a1869f9385/src/data/emacs-mode/agda-input.el#L516

rfindler:
It looks like the lowercase bb letters don't work with the current set of latex includes we use. That is, the `\mathbb` macro doesn't like them. [This stackexchange answer](https://tex.stackexchange.com/questions/254445/package-for-mathbblowercase-letter-symbol) suggests that we could add a new dependency on the latex generated output but that seems scary. So, for now, I think I'll just do the uppercase letters and the digits.

rfindler:
.... actually it doesn't work for the digits either! So I'll leave those out for now too. sigh.

rfindler:
Well, I just added them and just wrote an English explanation in the DrRacket manual. FWIW.

bdeket:
Thank you very much.

PR #326, opened 2024-04-25
cocoa,canvas: fix bezel-cell frame for FocusView

### LINKED ISSUE #325: macOS: text field misrendering
I've tested this using a fresh build of Racket HEAD on macOS 14.4 (arm64) and on the `release` branch. The following program

```racket
#lang racket/gui

(define frame
  (new frame%
       [label "Test"]))

(define msg
  (new message%
       [label "test"]
       [parent frame]))

(define input
  (new text-field%
       [label "test"]
       [parent frame]))

(send frame show #t)
```

renders as

Screenshot 2024-04-25 at 07 42 38

The background of the text field seems to extend past its boundary and cover other widgets. The problem is worse in more complex apps:

Screenshot 2024-04-25 at 07 44 57

Screenshot 2024-04-25 at 07 46 27

When I try a [recent snapshot build](https://users.cs.utah.edu/plt/snapshots/2024-04-23-7844684fb3/index.html), the bug doesn't happen, so the issue might be with some recent change to racket itself, since there haven't been recent commits here, though I haven't been able to pinpoint the commit.

Bogdanp:
Fixes #325. The `r` argument represents the portion of the view that is "dirty", and it seems like it may be larger than the view's frame[1]. It looks like this code was already intending to use the view's frame to draw the bezel, but accidentally used the dirty rect instead.

[1]: At least on recent versions of macOS, which may explain why [this isn't a problem when using versions of Racket built on older macOS versions](https://github.com/racket/drracket/issues/652#issuecomment-1853144763).

### DIFF
--- a/gui-lib/mred/private/wx/cocoa/canvas.rkt
+++ b/gui-lib/mred/private/wx/cocoa/canvas.rkt
@@ -145,16 +145,16 @@
   (tell (tell NSTextFieldCell alloc) initTextCell: #:type _NSString ""))
 (tellv bezel-cell setBezeled: #:type _BOOL #t)
 
-(define-objc-class FocusView NSView 
+(define-objc-class FocusView NSView
   [on?]
   (-a _void (setFocusState: [_BOOL is-on?])
       (set! on? is-on?))
   (-a #:async-apply (box (void))
       _void (drawRect: [_NSRect r])
       (let ([f (tell #:type _NSRect self frame)])
-        (tellv bezel-cell 
+        (tellv bezel-cell
                drawWithFrame: #:type _NSRect (make-NSRect (make-NSPoint 2 2)
-                                                          (let ([s (NSRect-size r)])
+                                                          (let ([s (NSRect-size f)])
                                                             (make-NSSize (- (NSSize-width s) 4)
                                                                          (- (NSSize-height s) 4))))
                inView: self))


mflatt:
Thanks!!

shhyou:
Did this get into Racket v8.13?

sorawee:
It probably doesn't matter anyway, right? The dmg package is unaffected. It affects only build from source, but in that case you can just pull a fixed version instead.

shhyou:
In our class, we have trouble running `big-bang` on macOS Sonoma where the `big-bang` window got incorrectly painted with the content of the Interaction Window. Using the snapshot build fixes it and I am wondering whether this issue is relevant.

mfelleisen:
A problem with `big-bang` in class would be highly relevant. 

- Is it reproducible? 
- Does it interfere with coding in class? 
- Can you upload a screenshot?

shhyou:
> A problem with `big-bang` in class would be highly relevant.
> 
>     * Is it reproducible?
> 

On the student's machine (macOS Sonoma, x86_64, Racket CS v8.12), the big-bang window is entirely blank and sometimes it gets painted with the content of the Interaction Window.

This happens consistently at least on that particular machine. I have not checked with others to see if this happens for other people.

>     * Does it interfere with coding in class?
> 

Sort of, because nothing ever shows up in the `big-bang` window.

>     * Can you upload a screenshot?

I'll try to find the student.

rfindler:
We're considering this as a fix that might need to go into the 8.13 release (which is halfway out the door) or if it can wait for 8.14 (in 3 months, give or take). Do folks have any opinion on this point? @jbclements

shhyou:
Actually, I realize that this is not urgent for v8.13. Assuming that v8.14 is going to be released in August, the fix can make it into Fall quarter/semester.

jbclements:
I'm seeing somewhat conflicting information about which processors & operating systems are affected. It appears that it's only macOS 14, Sonoma, but is it only x86_64 or does it affect ARM as well?

Also, I'm hoping someone can confirm that this bug was present in 8.12 as well?

The rather visible effect in the preference pane is distressing, to be sure.

mflatt:
On AArch64, I believe the bug affects only users who build Racket from source. It doesn't affect users who run a release or snapshot distribution. (It's not so much about a release/snapshot being built *on* an old version of the OS as being built *for* older versions of the OS by selecting the 11.1 SDK. I have seen this effect one other time: racket/racket@eccc500e6e56b7f06ea7843089b6df364a241c56.)

I am unclear on the status for x86_64. I have seen older release distributions of the x86_64 executable run a big-bang-like program badly on an AArch64 machine via Rosetta, and I had chalked that up to a Rosetta issue, but maybe it was the same bug. In any case, that's not the same as running on an x86_64 Sonoma installation.

shhyou:
Here is the code that's not working on the student's machine. With the code in the ZIP file, it should be showing a static, non-moving game. The same code with the snapshot build works.
[Exercise 5 Snake.zip](https://github.com/racket/gui/files/15263555/Exercise.5.Snake.zip)

sorawee:
@shhyou just to be clear, did the student build their Racket from source? Or ran Racket via Rosetta?

I don't think "the snapshot build works" tells anything, because the snapshot build has always been working fine.

shhyou:
No, the student uses v8.12 dmg and is on an x86_64 machine.

jbclements:
@shhyou I'm not sure how to download the code that you posted. I see the text "Exercise 5 Snake.zip (see edit history)", but I'm not sure how to interpret that, or to download the code. I'm also wondering whether it isn't sufficient to run the code that @Bogdanp provided in the bug report, to wit:

```racket
#lang racket/gui

(define frame
  (new frame%
       [label "Test"]))

(define msg
  (new message%
       [label "test"]
       [parent frame]))

(define input
  (new text-field%
       [label "test"]
       [parent frame]))

(send frame show #t)

```

jbclements:
Also, there seems to be some conflict between Matthew's suggestion that this does not affect snapshot or release builds (because they target the 11.1 SDK), and @shhyou 's evidence of a student that is apparently having this problem using the 8.12 release dmg.

FWIW, I just tested the release bundle on an x86_64 mac running macOS Monterey (12.7.4), and unsurprisingly, I don't see the problem.

jbclements:
Okay... I just tried to upgrade my x86_64 mac to Sonoma, and Sonoma simply isn't available for my (2015) x86_64 machine. Ah well.

shhyou:
> @shhyou I'm not sure how to download the code that you posted. I see the text "Exercise 5 Snake.zip (see edit history)", but I'm not sure how to interpret that, or to download the code.

Sent it offline, just in case.

rfindler:
I'd like to help test but I'm not sure if what I have is useful. It is a late 2013 imac running high sierra (10.13). It doesn't have the problem with the snake game in 8.12, on either of the 32 or 64 bit builds. I somehow have the (possibly incorrect?) impression that it might be a useful machine to test a fixed Racket on for the release. I'm happy to do so!

@jbclements there is a little popdown menu near the word "edited" in @shhyou 's message with the snake game; if you go to the first version of the message, you'll be able to download the game. At least, that worked for me.

rfindler:
> No, the student uses v8.12 dmg and is on an x86_64 machine.

I updated an x86_64 machine I have to sonoma and used the x8.12 dmg with the snake game code posted here earlier and I didn't see any problems. (I'm surprised by this ...)

PR #329, opened 2024-06-28
add a rectangle mode for indent guides

rfindler:
This just changes the drawing to see if these are the desired rectangles.


rfindler:
This is all very experimental and, once I could show it to @mflatt we understood fairly quickly that this isn't the right approach. Probably I shouldn't have opened a pull request. Too optimistic! 

I'll guess I'll keep using this pull request and when it gets to the point that neither of immediately rejects it, I'll try to explain. Does that sound okay?

rfindler:
Or maybe I should just stick with a branch. Actually, that's sounds better. Let me close the pull request.

Issue #330, opened 2024-07-06
User can't launch DrRacket due to problem in `key-translate`


soegaard:
From https://www.reddit.com/r/Racket/comments/1dwusoe/help_i_cant_use_the_drracket_app/

```
[lambduli](https://www.reddit.com/user/lambduli)
Hey everyone,

Today I installed the latest version `8.13` and it doesn't seem to work.

Here's what my setup is and what I tried:

I am on Apple MacBook Air with M1 chip. The latest OS.

The app from the launcher doesn't work. The icon bounces but then stops.

When I tried running it from the terminal I got this error:

integer->char: contract violation
  expected: valid-unicode-scalar-value?
  given: 55349
  context...:
   /Applications/Racket v8.13/share/pkgs/gui-lib/mred/private/wx/cocoa/key-translate.rkt:454:0: key-translate
   /Applications/Racket v8.13/share/pkgs/gui-lib/mred/private/wx/cocoa/key-translate.rkt:552:0
   body of "/Applications/Racket v8.13/share/pkgs/gui-lib/mred/private/wx/cocoa/key-translate.rkt"
   body of "/Applications/Racket v8.13/share/pkgs/gui-lib/mred/private/wx/platform.rkt"
   
   EDIT
   So it must be related to my layout. I have a custom keyboard layout with bunch of special symbols on it. When I switch to a simpler layout I can open DrRacket just fine. I think this might be a bug. I should probably open an issue.
   ```
 
 The last line of `key-translate` is:
 ```
 (integer->char (ptr-ref output-chars _UniChar i))
 ```
 
The number 55349 from the error message falls outside the ranges accepted by `integer->char`.
The accepted ranges are from 0 to D7FF and from E000 to 10FFFF.

The documentation from Apple says an array of UniChar is returned.
UniChar stands for "UTF-16 code units".

A search for 55349 and UTF 16 confirms that 55349 is not a valid UTF 16 character.

Suggestion: We return some dummy value like `#\space` for integers from D800 to DFFF.

lambduli:
Here is my keyboard layout that seems to be causing the issues.
[Code.keylayout.zip](https://github.com/user-attachments/files/16117409/Code.keylayout.zip)
I had to zip it for GitHub. I will add screenshots in the next one.

lambduli:
Screenshots of my layout.

No special keys pressed:
[...]

PR #331, opened 2024-07-08
Fix Issue #330

soegaard:
Ignore non-valid unicode characters.

### DIFF
--- a/gui-lib/mred/private/wx/cocoa/key-translate.rkt
+++ b/gui-lib/mred/private/wx/cocoa/key-translate.rkt
@@ -492,8 +492,10 @@
                   output-chars)
   ; get the number of characters returned, and convert to string
   (define n (max 0 (min max-string-length (unbox actual-string-length))))
-  (list->string (for/list ([i (in-range n)]) 
-                  (integer->char (ptr-ref output-chars _UniChar i)))))
+  (list->string (for/list ([i (in-range n)]
+                           #:do [(define c (ptr-ref output-chars _UniChar i))]
+                           #:when (not (<= #xD800 c #xDFFF)))
+                  (integer->char c))))
 
 ;;;
 ;;; Conversions back and forth between characters and key codes.


mflatt:
I imagine these values are generated for characters in supplementary planes. Should input handling track surrogate pairs and join them to form those characters?

lambduli:
@soegaard I have just tried out all the characters in my custom layout. Everything works exactly as I would expect.

As per dead keys — my custom layout doesn't have any. (But my standard Czech layout has two and those work just fine in DrRacket.)

soegaard:
@lambduli Thanks for the report.

@mflatt Since Jan is the only one that come across the problem, and his layout seems to work,
I am inclined to keep the PR as-is.

mflatt:
I don't think the bits described in the quoted documentation are the relevant ones. Those seem maybe part of a `uchr` resource.

Meanwhile, the result is a `UniChar` array, which is UTF-16, and characters like '𝔸' need surrogate encoding in UTF-16. I think the right change is

```
(define output-chars (malloc _UniChar (add1 max-string-length))) ; leave room for a terminator
```
and then
```
  ;; get the number of characters returned, and convert to string;
  ;; the characters are UTF-16
  (define n (max 0 (min max-string-length (unbox actual-string-length))))
  (ptr-set! output-chars _UniChar (add1 n) 0) ; ensure nul terminator
  (cast output-chars _pointer _string/utf-16))
```

soegaard:
Your solution certainly looks better.
I tested it and it works for me.

@lambduli Can you change this last version?

Around line 483 the code now looks like this:

```
  (UCKeyTranslate layout 
                  virtual-key-code
                  key-action
                  (bitwise-and (>> modifier-key-state 8) #xFF)
                  keyboard-type
                  key-translate-options
                  dead-key-state
                  max-string-length
                  actual-string-length
                  output-chars)
  ;; get the number of characters returned, and convert to string;
  ;; the characters are UTF-16
  (define n (max 0 (min max-string-length (unbox actual-string-length))))
  (ptr-set! output-chars _UniChar n 0) ; ensure nul terminator
  (cast output-chars _pointer _string/utf-16))
```

And in line 429 an `add1` was added:
```
(define output-chars (malloc _UniChar (add1 max-string-length))) ; leave room for a terminator
```

lambduli:
Hi @soegaard, I am sorry for such a long delay! I had to take time off because of some deadlines and such.

I am happy to try that change.
So I replaced the code that began at line 483 with the new one (rewriting the few lines that featured the original fix from #330).

I also replaced the 429th line.

I tried to run DrRacket and it launched fine it also seems that all my special characters work in the editor just fine.

mflatt:
Closing as fixed by #334.

PR #332, opened 2024-07-19
add a few test cases that cover the basics of the highlighting of invisible parens

rfindler:
This adds some test cases, but also a dependency on a package that's not dependedn on (or in) main-distribution-test. Is this okay, @samth (once the rhombus-prototype repo gets split up, I mean)?


rfindler:
@samth I'll go ahead and push this; we have an entire release cycle to figure out an alternative if you see a problem.

PR #337, opened 2024-09-15
Fix #336: Disable asynchronous mode for GTK printing.

### LINKED ISSUE #336: Drracket prints definitions as blank pages
This issue is related to https://github.com/racket/drracket/issues/465. On Linux, "File > Print definitions... > [print dialogue opens]" generates empty pages. This is tested with both Racket 8.13 and 8.14 on multiple Linux distro. And it seems the problem has existed for a while. 8.13 and 8.14 are not the only two versions that suffer from it. My tests show that disabling asynchronous operation in the following line fixes this problem. The reason for that is still not clear. But it should be safe for now to just disable asynchronous printing.
https://github.com/racket/gui/blob/03b7ffaab6ba1bd5b16907b95a1a0f12e9dd6135/gui-lib/mred/private/wx/gtk/printer-dc.rkt#L168

chihyang:
Fix issue #336: for unknown reasons, asynchronous printing in GTK3 generates blank pages.  Disabling asynchronous mode fixes this problem.  This change makes the main window freeze when the print dialog is popped up.  But it shouldn't be a problem.  First this is also the default behavior when the print dialog is popped up in Windows.  This change makes the behavior more consistent on different OSes.  Besides, it is reasonable to forbid editing when the contents of the editor is being printed.

### DIFF
--- a/gui-lib/mred/private/wx/gtk/printer-dc.rkt
+++ b/gui-lib/mred/private/wx/gtk/printer-dc.rkt
@@ -165,7 +165,7 @@
     (connect-end-print op-gtk)
 
     (gtk_print_operation_set_n_pages op-gtk (length pages))
-    (gtk_print_operation_set_allow_async op-gtk #t)
+    (gtk_print_operation_set_allow_async op-gtk #f)
     (gtk_print_operation_set_default_page_setup op-gtk page-setup)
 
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #341, opened 2024-10-30
Attempt to fix autofixer installation again

jackfirth:


.github/workflows/resyntax-autofixer.yml
```diff
@@ -23,8 +23,11 @@ jobs:
           dest: '"${HOME}/racketdist-current-CS"'
           sudo: never
       - name: Install and setup
-        run: raco pkg install --auto gui-doc gui-lib gui-test gui text-table
+        run: |
+          raco pkg install -i --auto --no-setup --skip-installed gui-test
+          raco pkg update --auto --no-setup gui-doc gui-lib gui text-table
+          raco setup gui-doc gui-lib gui-test gui text-table
```
samth, line 29: you might want -D here

jackfirth: I won't bother for now, since eventually I'd like Resyntax to work on scribble source files too. See jackfirth/resyntax#387 for an example use case.

PR #342, opened 2024-11-19
Fix autofixer version

jackfirth:
Accidentally used the wrong version number


sorawee:
@jackfirth this is the same error in https://github.com/racket/scribble/pull/459. Do you know what's going on?

jackfirth:
> @jackfirth this is the same error in https://github.com/racket/scribble/pull/459. Do you know what's going on?

I think I broke the GitHub review integration when I changed Resyntax to use multipass analysis. I'll fix that.

PR #349, opened 2025-03-30
docs: correct result contract on get-directory

benknoble:
Using path? generates a link to the predicate, like the result of get-file.


sorawee:
Thanks!

PR #358, opened 2025-08-31
queue: fix range of yield/no-sync

Bogdanp:
I noticed that `yield/no-sync` could potentially return `void?` in addition to `#t` or `#f`, which seems like a potential bug.

### DIFF
--- a/gui-lib/mred/private/wx/common/queue.rkt
+++ b/gui-lib/mred/private/wx/common/queue.rkt
@@ -495,24 +495,25 @@
             (wait-now))]))]))
 
 (define (yield/no-sync)
-  (let ([e (current-eventspace)])
-    (when (eq? (current-thread) (eventspace-handler-thread e))
-      (let ([v (sync/timeout 0 ((eventspace-queue-proc e) #f #f #f))])
-        (if v
-            (begin (handle-event v e) #t)
-            #f)))))
+  (define e (current-eventspace))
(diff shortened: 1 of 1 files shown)


mflatt:
Good catch, but I think the documentation promises a `#t` result instead of `#f` when the current thread is not the handler thread. I'm not sure that was the right choice, but it's more consistent with a `(void)` result, so probably we want to align with the documentation.

Bogdanp:
Note that this change is about `yield/no-sync`, which is private and undocumented (afaict). I was concerned about uses like https://github.com/racket/gui/blob/5e5dc582c4f156fd907411db0d785f20f7fe130a/gui-lib/mred/private/wx/cocoa/slider.rkt#L34 where a void result would cause an infinite loop.

My reading of the documentation for `yield` is that a `#f` result is expected in this case (a `yield` call without a positional arg):

> If no argument is provided, [yield](dfile:///Users/bogdan/Library/Application%20Support/Dash/DocSets/Racket/Racket.docset/Contents/Resources/Documents/docs.racket-lang.org/gui/Windowing_Functions.html#%28def._%28%28lib._mred%2Fmain..rkt%29._yield%29%29) dispatches an unspecified number of events, but only if the current thread is the current eventspace’s handler thread (otherwise, there is no effect). The result is #t if any events may have been handled, #f otherwise.

I could definitely be missing something, though.

PR #369, opened 2026-06-17
mrlib/gif: don't depend on racket/gui/base

AlexKnauth:
The only identifier it had been importing from `racket/gui/base` was `bitmap%`, which can be imported from `racket/draw` instead.


mflatt:
Thanks!

PR #371, opened 2026-09-20
font dialog: keep sample text white in dark mode

otherjoel:
Fixes an issue where, in dark mode on macOS, the sample text in the `get-font-from-user` dialog can't be seen. It is black text on a black field.

This issue was found and patch written by Claude:

> `text-field%` makes its text white in dark mode. But the dialog replaces the sample's style each time a control changes, and once before it opens, so the white color is lost.

It looks like the right fix, and I can verify it works on macOS Sequioa in dark mode with Racket 9.3. But scrutinize accordingly.


rfindler:
This looks reasonable to me.

